PR #7, opened 2023-09-14
Refactor beam_center_finder to use Sciline

SimonHeybrock:
~Note that this depends on a bugfix to be made in Sciline, since `ScopeTwoParams` can not be used as a parameter, currently.~ Sciline was updated.


src/esssans/beam_center_finder.py
```diff
@@ -227,15 +229,11 @@ def minimize(
...
-    wavelength_bins: sc.Variable,
-    q_bins: Union[int, sc.Variable],
-    gravity: bool = False,
-    minimizer: str = 'Nelder-Mead',
-    tolerance: float = 0.1,
-) -> Tuple[sc.Variable, sc.Variable]:
+    data: MaskedData[SampleRun],
+    graph: ElasticCoordTransformGraph,
+    wavelength_bins: WavelengthBins,
+    norm: Clean[SampleRun, Denominator],
```
SimonHeybrock, line 235: @nvaytet By using this, we will automatically get the denominator including the direct-beam factor, which was not used in the old implementation. If I remember correctly the argument was that we need the beam-center-find for the direct-beam iteration. However, `DirectBeam` is an optional dependency in the graph, so I think this should just work...

... unless there is an explicit reason why we should *not* include the direct-beam factor here?

nvaytet: > we need the beam-center-find for the direct-beam iteration

correct

src/esssans/beam_center_finder.py
```diff
@@ -102,28 +109,20 @@ def iofq_in_quadrants(
...
-            q_bins=q_bins,
-            gravity=gravity,
-            wavelength_bands=wavelength_range,
-        )
-        # Normalize
-        out[quad] = normalize(numerator=data_q, denominator=norm_q).hist()
+        params[CleanMasked[SampleRun, Numerator]] = data[sel]
+        params[CleanMasked[SampleRun, Denominator]] = norm[sel]
+        pipeline = sciline.Pipeline(providers, params=params)
+        out[quad] = pipeline.compute(IofQ[SampleRun])
```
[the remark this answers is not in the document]
SimonHeybrock: No, it is not a chicken-egg problem: The outer pipeline uses the provider for `BeamCenter`, which depends on certain things that can be computed without knowing the beam center. Internally, this provider defines a *different* pipeline, used to compute `IofQ`. Note that we set up this pipeline without file-loading etc., but just set the known inputs as parameters.

src/esssans/beam_center_finder.py
SimonHeybrock: ... but it is fine to use it if we have it?

nvaytet: Yes, and I would actually argue that if it is pixel dependent, then it is probably required?
I think we dropped it in the sans2d case because we assumed the direct beam was just wavelength dependent, and we were trying to overlap 4 curves. Including/omitting the direct beam would affect the shape of all 4 curves in the same manner, and we can thus drop it (I hope I didn't mess something up in my logic).

However, if it is now pixel dependent, then probably we cannot since different quadrants will be affected differently?

Maybe this is our chicken-egg problem then, because the direct beam computation requires the beam center finding and vice versa?

SimonHeybrock: You can still run it without. The `BeamCenter` is optional in the `esssans` pipeline, just as it was optional in `ess.sans`.

nvaytet: I think I wasn't really asking whether you can run the pipeline or not, but more conceptually, how are we going to compute the direct beam in practise?

Even without thinking about `sciline`, if you want to compute a pixel-dependent direct beam, you need to find the beam center. But to compute that, you need the direct beam. Do you just have a first guess for the beam center?

SimonHeybrock: > Even without thinking about sciline, if you want to compute a pixel-dependent direct beam, you need to find the beam center. But to compute that, you need the direct beam. Do you just have a first guess for the beam center?

Didn't you prove with your work on `ess.sans.beam_center_finter` that the direct beam is *not* required for finding the beam center?

nvaytet: That was my conclusion, as long as the direct beam is not pixel-dependent...

nvaytet (approved):

PR #11, opened 2023-09-15
Fix flaky result caused by modification of input coord

### LINKED ISSUE #10: Flaky result values?
After implementing upper-bound variances, it was noticed that there was a small discrepancy between the *values* in the two approaches:

Screenshot 2023-09-11 at 10 23 57

While trying to track this down, I have the impression that there is a small "flakyness" in the values: My unit test comparing the results of the two approaches passes sometimes, but fails other times.

Is this just multi-threading-related floating point rounding errors, or is there a bigger issues?

SimonHeybrock:
This problem only shows up with the DaskScheduler.

Fixes #10.

Note that I probably introduced this bug during the refactor to Sciline, i.e., is was not present in `ess.sans`.


SimonHeybrock:
I have added `dask` as a package dependency. This may not be what we want to use in production in the long run, but for now Sciline's `NaiveScheduler` is not a good alternative: It keeps all intermediate results and may thus result in excessive memory consumption.

PR #14, opened 2023-09-18
Fix application of beam center

### LINKED ISSUE #13: Beam center offset not applied to normalization
This was broken in a flaky manner before #11 (but usually worked, thanks to a side effect), but after #11 it is consistently broken.

Should be fixed by changing graph `MaskedData->Clean` to `MaskedData->CalibratedMaskedData->Clean`, i.e., extract the beam-center offset from the `detector_to_wavelength` function. Then change solid angle computation to depend on `CalibratedMaskedData`

Subtlety: Figure out how to avoid cycle when using beam-center finder.

SimonHeybrock:
Fixes #13.

Also re-compute solid angle during iteration. This change resulted in a ~3mm difference in the result for the beam center, compared to the pre-Sciline implementation.

### DIFF
--- a/src/esssans/beam_center_finder.py
+++ b/src/esssans/beam_center_finder.py
@@ -9,37 +9,37 @@ import scipp as sc
 
 from .conversions import (
     ElasticCoordTransformGraph,
-    to_Q,
-    mask_wavelength,
     calibrate_positions,
     detector_to_wavelength,
+    mask_wavelength,
+    to_Q,
 )
 from .i_of_q import merge_spectra
 from .logging import get_logger
 from .normalization import (
-    normalize,
     iofq_denominator,
+    normalize,
     solid_angle_rectangular_approximation,
 )
 from .types import (
     BeamCenter,
     Clean,
+    CleanDirectBeam,
     CleanMasked,
+    CleanMonitor,
     Denominator,
+    DirectRun,
+    Incident,
     IofQ,
     MaskedData,
     Numerator,
     QBins,
     SampleRun,
+    SolidAngle,
+    Transmission,
+    UncertaintyBroadcastMode,
     WavelengthBands,
     WavelengthBins,
-    CleanMonitor,
-    Transmission,
-    DirectRun,
-    Incident,
-    SolidAngle,
-    CleanDirectBeam,
-    UncertaintyBroadcastMode,
 )
 
 
--- a/src/esssans/beam_center_finder.py
+++ b/src/esssans/beam_center_finder.py
@@ -102,9 +102,6 @@ def iofq_in_quadrants(
         The quadrants are named 'south-west', 'south-east', 'north-east', and
         'north-west'.
     """
-    # Offset the position according to the input shift
-    center = _offsets_to_vector(data=data, xy=xy, graph=graph)
-
     pi = sc.constants.pi.value
     phi = data.transform_coords(
         'phi', graph=graph, keep_intermediate=False, keep_inputs=False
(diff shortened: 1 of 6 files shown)


src/esssans/beam_center_finder.py
```diff
@@ -94,33 +102,38 @@ def iofq_in_quadrants(
...
+        detector_to_wavelength,
+        calibrate_positions,
+        solid_angle_rectangular_approximation,
+    ]
     params = {}
+    params[UncertaintyBroadcastMode] = UncertaintyBroadcastMode.upper_bound
     params[WavelengthBands] = wavelength_range
     params[QBins] = q_bins
     params[ElasticCoordTransformGraph] = graph
+    params[BeamCenter] = _offsets_to_vector(data=data, xy=xy, graph=graph)
```
SimonHeybrock, line 127: This shows an interesting "shortcoming" of Sciline: We want some specific subset of providers, basically something that is obtained by replacing sub-branches with concrete parameters. Is this indicating that we need a different interface, or helper?

Basically we want to set a parameter, and it should drop all providers that are directly or indirectly used to compute this?

Or maybe what we have here is fine, since it is explicit?

nvaytet: Not sure I really understand the issue here. Isn't what you have fine?

src/esssans/conversions.py
```diff
@@ -157,19 +158,29 @@ def monitor_to_wavelength(
     )
 
 
+def calibrate_positions(
```
nvaytet, line 161: :+1:

src/esssans/normalization.py
```diff
@@ -102,14 +103,20 @@ def transmission_fraction(
...
     data_transmission_monitor: CleanMonitor[RunType, Transmission],
     direct_incident_monitor: CleanMonitor[DirectRun, Incident],
     direct_transmission_monitor: CleanMonitor[DirectRun, Transmission],
-    solid_angle: SolidAngle[RunType],
     direct_beam: Optional[CleanDirectBeam],
     uncertainties: UncertaintyBroadcastMode,
-) -> Clean[RunType, Denominator]:
+) -> NormWavelengthTerm[RunType]:
     """
     Compute the denominator term for the I(Q) normalization. This is basically:
```
nvaytet, line 121: So now you have `iofq_norm_wavelength_term` and `iofq_denominator`. Is this docstring more for the latter function?

src/esssans/normalization.py
```diff
@@ -152,17 +159,22 @@ def iofq_denominator(
...
     # The direct beam may also be pixel-dependent. In this case we need to drop
     # variances of the other term already before multiplying by solid_angle.
-    broadcast = {
-        UncertaintyBroadcastMode.drop: drop_variances_if_broadcast,
-        UncertaintyBroadcastMode.upper_bound: broadcast_with_upper_bound_variances,
-        UncertaintyBroadcastMode.fail: lambda x, sizes: x,
-    }[uncertainties]
+    broadcast = _broadcasters[uncertainties]
     if direct_beam is not None:
         denominator = direct_beam * broadcast(denominator, sizes=direct_beam.sizes)
```
nvaytet, line 164:
```suggestion
    if direct_beam is not None:
        denominator = direct_beam * _broadcasters[uncertainties](denominator, sizes=direct_beam.sizes)
```

src/esssans/beam_center_finder.py
SimonHeybrock: It is indeed fine, just slightly verbose, and easy to forget updating if we even change the main workflow.

src/esssans/normalization.py
SimonHeybrock: Not sure I see the difference? If it is just style, it will probably be longer or harder to read since black will add weird linebreaks.

src/esssans/normalization.py
SimonHeybrock: Thanks, see update!

src/esssans/normalization.py
nvaytet: Just thought it removed the need to create the `broadcast` variable outside of the `if direct_beam is not None:`.
Could also be
```Py
    if direct_beam is not None:
        broadcast = _broadcasters[uncertainties]
        denominator = direct_beam * broadcast(denominator, sizes=direct_beam.sizes)
```
?

I really don't mind, it was a suggestion. As you say, it's mostly style.

Sometimes I wish github had two types of suggestions: one where you actually would like to see the change made, and another which would be more optional, an actual "suggestion", as in "take it or leave it"

SimonHeybrock: In Python this does not make a difference anyway, this is not C(++) scoping rules.

SimonHeybrock:
@nvaytet Do you have more comments on this?

src/esssans/normalization.py
```diff
@@ -152,17 +161,51 @@ def iofq_denominator(
...
+    Compute the denominator term for the I(Q) normalization.
+
+    This is basically:
+    ``solid_angle * direct_beam * data_transmission_monitor * direct_incident_monitor / direct_transmission_monitor``
+    The `wavelength_term` included all but the `solid_angle` and is computed by
+    :py:func:`iofq_norm_wavelength_term`.
+
+    Because the multiplication between the wavelength dependent terms (monitor counts)
+    and the pixel dependent term (solid angle) consists of a broadcast operation which
+    would introduce correlations, we strip the data of variances.
```
nvaytet, line 188: > we strip the data of variances.

I guess this is not always true, right? Since there is a mode for computing an upper bound?

SimonHeybrock: Yes, I updated it. The reference is not in the docstring of `UncertaintyBroadcastMode`, so I removed it here.

nvaytet (approved):

PR #16, opened 2023-10-03
Fix center-of-mass estimation of beam center

SimonHeybrock:
- Fix ignored masks
- Make approach viable, by ignoring low-count regions that lead to asymmetric cutoffs, when the scattering patterns touches the edges of the detector panel.

I have enabled this as a default in the workflow, since it takes less than a second (as opposed to 20 seconds).

### DIFF
--- a/src/esssans/beam_center_finder.py
+++ b/src/esssans/beam_center_finder.py
@@ -6,6 +6,7 @@ from typing import Dict, List, NewType, Optional, Union
 import numpy as np
 import sciline
 import scipp as sc
+from scipp.core import concepts
 
 from .conversions import (
     ElasticCoordTransformGraph,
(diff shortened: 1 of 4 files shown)


src/esssans/beam_center_finder.py
```diff
@@ -34,24 +35,68 @@
...
+) -> BeamCenter:
     """
-    Find the center-of-mass of the data counts.
+    Estimate the beam center via the center-of-mass of the data counts.
+
     The center-of-mass is simply the weighted mean of the positions.
+    Areas with low counts are excluded from the center of mass calculation, as they
+    typically fall into asymmetric regions of the detector panel and would thus lead
+    to a biased result. The beam is assumed to be roughly aligned with the Z axis.
+    The returned beam center is the component normal to the bema direction, projected
```
nvaytet, line 57:
```suggestion
    The returned beam center is the component normal to the beam direction, projected
```

src/esssans/beam_center_finder.py
```diff
@@ -34,24 +35,68 @@
...
+    onto the X-Y plane.
 
     Parameters
     ----------
     data:
-        The data to find the center-of-mass of.
+        The data to find the beam center of.
+    graph:
```
nvaytet, line 64: Should the graph be an input argument? Seems a bit strange to feed in the entire graph here, with Q and two_theta, when all we want is the `incident_beam`?
Maybe we should break up the sans graph into smaller pieces, and import it inside the function?

nvaytet: > Wouldn't be against everything we wanted to avoid with Sciline?

Yes, I guess you are right

nvaytet (approved):

PR #18, opened 2023-10-04
Add basics API reference

SimonHeybrock:
This is not 100% complete as we are still working on a more user-friendly solution in some cases, but these are the most important bits.

### DIFF
--- a/docs/api-reference/index.md
+++ b/docs/api-reference/index.md
@@ -1,0 +1,29 @@
+# API Reference
+
+## Classes
+
+```{eval-rst}
+.. currentmodule:: {{projectname}}
+
+.. autosummary::
+   :toctree: ../generated/classes
+   :template: class-template.rst
+   :recursive:
+```
+
+## Top-level functions
+
+```{eval-rst}
+.. autosummary::
+   :toctree: ../generated/functions
+   :recursive:
+```
+
+## Submodules
+
+```{eval-rst}
+.. autosummary::
+   :toctree: ../generated/modules
+   :template: module-template.rst
+   :recursive:
+```
--- a/docs/api-reference/index.md
+++ b/docs/api-reference/index.md
@@ -3,7 +3,7 @@
 ## Classes
 
 ```{eval-rst}
-.. currentmodule:: {{projectname}}
+.. currentmodule:: esssans
 
 .. autosummary::
    :toctree: ../generated/classes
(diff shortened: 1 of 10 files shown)


docs/api-reference/index.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Submodules
+
+```{eval-rst}
+.. autosummary::
+   :toctree: ../generated/modules
+   :template: module-template.rst
+   :recursive:
+
+   beam_center_finder
```
nvaytet, line 32: Is there a way we can improve the long function names? Maybe only have the function name and not the submodule path? But I guess that path is important information to have...
Screenshot at 2023-10-04 15-16-28

docs/api-reference/index.md
```diff
@@ -0,0 +1,35 @@
...
+   sans2d
+   types
```
nvaytet, line 34: Was it intentional that some entries under "Module attributes" and others under "Classes"? I don't really seem to get the logic separating the two sections.

docs/api-reference/index.md
SimonHeybrock: I suppose we can shorten, `beam_center_finder` -> `beam_center`, or remove the duplicate `beam_center_` prefix from the 2 finders?

docs/api-reference/index.md
SimonHeybrock: It is not intentional, but this is what we get from the generic template, and due to the fact that some domain types are using `NewType` while others are not (are actual classes). I don't know how to fix that. Maybe the template could be rewritten?

docs/api-reference/index.md
nvaytet: > remove the duplicate beam_center_ prefix from the 2 finders?

would work in this case. But I guess it won't save us from future functions with long names.

SimonHeybrock: Maybe having shorter module names is good style generally. Should we just rename `beam_center_finder` -> `beam`?

After all, PEP8 says: "Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability."

nvaytet: Or `beamcenter`? I feel `beam` is a little too vague?

SimonHeybrock: I suggest we keep it for now. We should probably sit down with Wojciech to review all the naming in the repo.

nvaytet (approved):

PR #21, opened 2023-10-16
Run copier update

SimonHeybrock:


SimonHeybrock:
Release build also passed: https://github.com/scipp/esssans/actions/runs/6529042215

PR #23, opened 2023-10-17
Run copier update and tox -e deps

SimonHeybrock:
Based on latest copier_template@nightly2.


requirements/nightly.txt
```diff
@@ -0,0 +1,173 @@
...
+sciline @ git+https://github.com/scipp/sciline@main
+    # via -r nightly.in
+scipp==23.8.0
+    # via
+    #   -r nightly.in
+    #   scippneutron
+    #   scippnexus
+scippneutron @ git+https://github.com/scipp/scippneutron@main
+    # via -r nightly.in
+scippnexus==23.8.0
```
nvaytet, line 136: Do we need to also get `scippnexus` from `git`?

SimonHeybrock: Hmm, good point, we need to allow for listing extra deps that are transitive.

SimonHeybrock: See update in scipp/copier_template#43

requirements/make_base.py
```diff
@@ -0,0 +1,44 @@
...
+with open("base.in", "w") as f:
+    f.write(header)
+    f.write("\n".join(dependencies))
+
+
+def as_nightly(name: str) -> str:
+    if name == "scipp":
+        version = f"cp{sys.version_info.major}{sys.version_info.minor}"
+        base = "https://github.com/scipp/scipp/releases/download/nightly/scipp-nightly"
+        suffix = "manylinux_2_17_x86_64.manylinux2014_x86_64.whl"
```
nvaytet, line 33: How are we making sure the `2_17` and `2014` numbers in the suffix are correct? I can see they come from the copier template, but they are hardcoded there as well. Do they depend on the version of `cibuildwheel` we are using to build scipp?

SimonHeybrock: > Do they depend on the version of cibuildwheel we are using to build scipp?

Yes, essentially. We have never changed this since the beginning of our wheel builds, so it will be a really rare thing. I do not want to remove it, since we may later also want to add mac, windows, ...

SimonHeybrock: Well, they might actually have changed, but I think the manylinux2014 was already there when we started, and the 2_17 is even older, iirc.

nvaytet (approved):

PR #25, opened 2023-10-24
Update HDF5 to avoid scipp.DataArray.attrs deprecation warnings

SimonHeybrock:
The new files are in `ess/loki/2`. I am currently not planning on changing any code in `ess`.


nvaytet (approved):
Did you check that the results were identical to that when using the old files?

SimonHeybrock:
It is not, and I do not understand why. Have there been other changes to the file?
image

SimonHeybrock:
I pushed an update to fix the discrepancy. I decided to not alter the files (since then the info is kind of lost) and instead applied the offsets in code. I did not apply a bench pos and beam-center offset. Result:

image

PR #27, opened 2023-11-02
Run copier update, adding unpinned nightly build

SimonHeybrock:
Note that the new build cannot really be tested before making a release.


SimonHeybrock:
I will merge this for testing purposes. If it works then https://github.com/scipp/copier_template/pull/67 represents the corresponding change in the template.

YooSunYoung (approved):

PR #31, opened 2023-11-15
Loki direct beam iterations

nvaytet:
This adds the necessary functionality to the I(Q) providers to compute the direct beam function for the LoKI detector tubes, using detector test data obtained at Larmor.

Changes to the pipeline:

- Ability to merge data from multiple runs using a parameter table
- `DirectRun` has been renamed to `EmptyBeamRun`
- Masking both before and after detector position calibration is now possible
- Normalization of detector and monitor counts using the sum of the indicent monitor counts are done for all runs

Computing the direct beam function is illustrated in a notebook.

Draft PR for now as unit tests are still lacking.


nvaytet:
Closing in favour of #35

PR #32, opened 2023-11-17
Save results to NXcanSAS

### LINKED ISSUE #15: Save final result to NXcansas format
What was implemented in https://github.com/scipp/ess/pull/190 should be migrated over to the workflow here.

jl-wynen:
Fixes #15


src/esssans/__init__.py
```diff
@@ -9,9 +9,25 @@
...
+    conversions,
+    i_of_q,
+    io,
+    metadata,
+    normalization,
+    sans2d,
+)
 
-providers = conversions.providers + i_of_q.providers + normalization.providers
+providers = (
```
jl-wynen, line 23: I changed providers to tuples everywhere in order to prevent accidental modifications by users. It is tempting to use code like `sans.providers.append`. But this leads to type errors because Sciline wants providers to be passed as a list. IMHO, this should be changed in Sciline, it should accept any sequence.

src/esssans/metadata.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def run_number(raw_data: RawData[SampleRun]) -> RunNumber:
+    """Get the run number from the raw sample data."""
+    return RunNumber(int(raw_data['run_number']))
+
+
+def run_title(raw_data: RawData[SampleRun]) -> RunTitle:
+    """Get the run title from the raw sample data."""
+    return RunTitle(raw_data['run_title'].value)
```
SimonHeybrock, line 14: I think those names a Mantid-specific, so maybe this should move into `sans2d`. NeXus has this information as well, but probably under slightly different names (I think `title` and `entry_identifier` in NXentry).

src/esssans/io.py
```diff
@@ -0,0 +1,24 @@
...
+import scipp as sc
+import scippnexus as snx
+from scippnexus.application_definitions import nxcansas
+
+from .types import BackgroundSubtractedIofQ, OutFilename, RunNumber, RunTitle
+
+
+def save_background_subtracted_iofq(
+    *,
+    iofq: BackgroundSubtractedIofQ,
```
SimonHeybrock, line 13: Do we ever need to store the result without background subtraction?

SimonHeybrock (commented):
Looks good, two small comments.

src/esssans/io.py
jl-wynen: Don't know. Given how Sciline works, we would need two separate functions. We can easily add another one if we need to.

src/esssans/metadata.py
jl-wynen: Good point. I'll move it.

SimonHeybrock (approved):

PR #34, opened 2023-11-20
Use transmission fraction in normalization

nvaytet:
It was realized in #31 that the incident monitor in the denominator is not necessarily from the same run at the incident monitor used in the transmission fraction (the transmission fraction is commonly computed from a 'transmission run').

We re-write the pipeline to take this into account.

### DIFF
--- a/src/esssans/normalization.py
+++ b/src/esssans/normalization.py
@@ -8,6 +8,8 @@ import scippneutron as scn
 from scipp.core import concepts
 
 from .types import (
+    BackgroundRun,
+    BackgroundTransmissionRun,
     CalibratedMaskedData,
     Clean,
     CleanDirectBeam,
@@ -22,6 +24,7 @@ from .types import (
     RawData,
     RawMonitor,
     RunType,
+    SampleRun,
     SampleTransmissionRun,
     SolidAngle,
     Transmission,
@@ -118,7 +121,7 @@ def transmission_fraction(
     frac = (sample_transmission_monitor / direct_transmission_monitor) * (
         direct_incident_monitor / sample_incident_monitor
     )
-    return TransmissionFraction(frac)
+    return TransmissionFraction[TransmissionRunType](frac)
 
 
 _broadcasters = {
@@ -177,16 +180,14 @@ def transmission_fraction_times_direct_beam(
     # Convert wavelength coordinate to midpoints for future histogramming
     # if wavelength_to_midpoints:
     out.coords['wavelength'] = sc.midpoints(out.coords['wavelength'])
-    return TransmissionFractionTimesDirectBeam[RunType](out)
+    return TransmissionFractionTimesDirectBeam[TransmissionRunType](out)
 
 
-def iofq_denominator(
-    transmission_fraction_times_direct_beam: TransmissionFractionTimesDirectBeam[
-        TransmissionRunType
-    ],
-    solid_angle: SolidAngle[RunType],
+def _iofq_denominator(
+    transmission_fraction_times_direct_beam: sc.DataArray,
+    solid_angle: sc.DataArray,
     uncertainties: UncertaintyBroadcastMode,
-) -> Clean[RunType, Denominator]:
+) -> sc.DataArray:
     """
     Compute the denominator term for the I(Q) normalization.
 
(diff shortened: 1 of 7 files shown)


src/esssans/normalization.py
```diff
@@ -110,63 +112,96 @@ def transmission_fraction(
...
+
+def iofq_norm_wavelength_term_sample(
+    incident_monitor: CleanMonitor[SampleRun, Incident],
+    transmission_fraction: TransmissionFraction[SampleTransmissionRun],
+    direct_beam: Optional[CleanDirectBeam],
+    uncertainties: UncertaintyBroadcastMode,
+) -> NormWavelengthTerm[SampleRun]:
+    """
+    Compute the wavelength-dependent contribution to the denominator term for the I(Q)
+    normalization, for the sample run.
```
nvaytet, line 175: Not sure what to do about the docstring here. Do we just duplicate between `iofq_norm_wavelength_term_sample` and `iofq_norm_wavelength_term_background`?

SimonHeybrock: So your problem is that `TransmissionFraction` depends on `[Sample|Background]TransmissionRun`, and you have this duplication (and the helper function above) do get away from that?

I feel this should be handled somehow when *computing* `TransmissionFraction`, not when using it.

Would a nested generic help?

```python
def transmission_fraction(
    sample_incident_monitor: CleanMonitor[Transmission[RunType], Incident],
    sample_transmission_monitor: CleanMonitor[Transmission[RunType], Transmission],
    direct_incident_monitor: CleanMonitor[DirectRun, Incident],
    direct_transmission_monitor: CleanMonitor[DirectRun, Transmission],
) -> TransmissionFraction[RunType]:

where `RunType` would not be `SampleRun` or `BackgroundRun`?

Comment (SimonHeybrock) [reply src/esssans/normalization.py:175]: To put it differently:
- The IofQ computation should not have to know how the transmission fraction was computed. The transmission fraction is the transmission fraction, and should not carry the difference between `SampleRun` and `SampleTransmissionRun` forward into the rest of the workflow.
- Splitting providers that do the same thing will lead to misleading graphs: Readers will think the operations do different things, so we should avoid this as much as possible.

File: src/esssans/sans2d.py [hunk 2]
```diff
@@ -58,27 +60,36 @@ def pooch_load_direct_beam(filename: DirectBeamFilename) -> DirectBeam:
     return DirectBeam(sc.io.load_hdf5(filename=get_path(filename)))
 
 
+def get_detector_data(
+    dg: LoadedFileContents[RunType],
+) -> RawData[RunType]:
+    return RawData[RunType](dg['data'])
```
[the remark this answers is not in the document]
nvaytet: Could be useful, yes

src/esssans/normalization.py
```diff
@@ -111,62 +109,58 @@ def transmission_fraction(
...
-    # The direct beam may also be pixel-dependent. In this case we need to drop
-    # variances of the other term already before multiplying by solid_angle.
+    """
+    out = incident_monitor * transmission_fraction
     if direct_beam is not None:
         broadcast = _broadcasters[uncertainties]
-        denominator = direct_beam * broadcast(denominator, sizes=direct_beam.sizes)
+        # TODO: Do we need an additional check for the case where the direct beam
+        # could be bin centers and the transmission fraction is bin edges? In that case
+        # we would want to raise instead of broadcasting.
```
SimonHeybrock, line 159: It would raise, because the broadcast will fail. Not sure the TODO is useful?

src/esssans/types.py
```diff
@@ -69,6 +76,11 @@
 BeamCenter = NewType('BeamCenter', sc.Variable)
 """Beam center, may be set directly or computed using beam-center finder"""
 
+LowCountThreshold = NewType('LowCountThreshold', sc.Variable)
+"""Threshold below which detector pixels should be masked
+(low-counts on the edges of the detector panel, and the beam stop)"""
```
SimonHeybrock, line 81: This mask seemed specific to Sans2d, what was the motivation for adding it in a generic place?

nvaytet: This needs to be set as a parameter in the notebook.
I think it would be annoying for the user to have to know that for this specific type, they have to do
```Py
params[sans.sans2d.LowCountThreshold] = sc.scalar(100, unit='counts')
```
but just e.g. `SampleRun` for all the other types.

So I am thinking of moving all the sans2d stuff into a folder, and making a `types.py` file in there, where we can then do
```Py
from esssans.types import *
from esssans.sans2d.types import *

params[LowCountThreshold] = sc.scalar(100, unit='counts')
```

I think after adding the loki stuff for the direct beam I wanted to move instrument specifics into subfolders anyway. Does that sound reasonable?

SimonHeybrock: I think scoping things is good, not bad. If a param is not generic for a technique then it would be confusing to have it on the "global" level. `import *` may be a antipattern, even here.

I would leave it for now, we can think about this more when adding support for actual ESS instruments.

SimonHeybrock: Isn't `sans.sans2d.LowCountThreshold` easier to remember than `sans.sans2d.types.LowCounthreshold`?

I am not at all convinced that having everything in `types` is ideal. Maybe all common params should be on the top level (technique or instrument), details in their submodule (example: reflectometry supermirror calibration with corresponding set of params).

SimonHeybrock (approved):

SimonHeybrock (approved):

PR #35, opened 2023-11-24
Loki Direct Beam Iterations (version 2)

nvaytet:
This adds the necessary functionality to the I(Q) providers to compute the direct beam function for the LoKI detector tubes, using detector test data obtained at Larmor.

Changes to the pipeline:

- Ability to merge data from multiple runs in the nexus file loader.
- `DirectRun` has been renamed to `EmptyBeamRun`
- A `DimsToKeep` parameter can be used to compute $I(Q)$ per detector layer.

Computing the direct beam function is illustrated in a notebook.


nvaytet:
I think the CI is failing because of #42. Any ideas on how to fix?

nvaytet:
Hmm something is wrong, the tests for beam center finder are taking almost 3 times as long as on `main`...

SimonHeybrock:
> I think the CI is failing because of #42. Any ideas on how to fix?

It hangs for me locally as well.

src/esssans/loki/general.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def detector_lab_frame_transform(
+    dg: LoadedFileContents[RunType],
+    instrument_path: NexusInstrumentPath,
+    detector_name: NexusDetectorName,
+    transform_path: TransformationChainPath,
+) -> LabFrameTransform[RunType]:
+    return LabFrameTransform[RunType](
+        dg[instrument_path][detector_name][transform_path]
```
SimonHeybrock, line 61: Do you have a plan for handling multiple detector banks, where all this will be indepdendent?

src/esssans/loki/masking.py
```diff
@@ -0,0 +1,117 @@
...
+    da = da.copy(deep=False)
+    counts = da.sum('tof').data
+    r = sc.sqrt(
+        da.coords['position'].fields.x ** 2 + da.coords['position'].fields.y ** 2
+    )
+    da.masks['low_counts_middle'] = (counts < sc.scalar(20.0, unit='counts')) & (
+        r < sc.scalar(0.075, unit='m')
+    )
+    # Low counts
+    da.masks['very_low_counts'] = counts < sc.scalar(3.0, unit='counts')
```
SimonHeybrock, line 75: Are you sure about this? I don't think we can just mask pixels with low counts, since they are very important for larger Q?

src/esssans/loki/masking.py
```diff
@@ -0,0 +1,117 @@
...
+        Raw data.
+    """
+    # Beam stop
+    da = da.copy(deep=False)
+    counts = da.sum('tof').data
+    r = sc.sqrt(
+        da.coords['position'].fields.x ** 2 + da.coords['position'].fields.y ** 2
+    )
+    da.masks['low_counts_middle'] = (counts < sc.scalar(20.0, unit='counts')) & (
+        r < sc.scalar(0.075, unit='m')
```
SimonHeybrock, line 72: This is before calibration or applying the beam center offset? What if the beam center is offset by more than the 7.5 cm (or a significant fraction of that)?

src/esssans/loki/masking.py
```diff
@@ -0,0 +1,117 @@
...
+        Mask for straws with low counts.
+    beam_stop_mask:
+        Mask for beam stop.
+    tube_edge_mask:
+        Mask for tube edges.
+    """
+    da = da.copy(deep=False)
+    # Clear masks from beam center finding step, as they are potentially using harsh
+    # thresholds which could remove some of the interesting signal.
+    da.masks.clear()
```
SimonHeybrock, line 101: Why do masks from beam-center finding show up in the actual data? This seems very brittle.

src/esssans/types.py
```diff
@@ -90,12 +111,25 @@ class NeXusMonitorName(sciline.Scope[MonitorType, str], str):
...
+class SamplePosition(sciline.Scope[RunType, sc.Variable], sc.Variable):
+    """Sample position"""
+
+
+class SourcePosition(sciline.Scope[RunType, sc.Variable], sc.Variable):
+    """Source position"""
+
+
+class DataWithLogicalDims(sciline.Scope[RunType, sc.DataArray], sc.DataArray):
+    """Raw data reshaped to have logical dimensions"""
```
SimonHeybrock, line 130: Does this need a name? Can't we do this directly in the loader?

src/esssans/direct_beam.py
```diff
@@ -0,0 +1,130 @@
...
+     3. Multiply the direct beam values inside each wavelength band by this factor
+     4. Compare the full-range $I(Q)$ to a theoretical reference and add the
+        corresponding additional scaling to the direct beam function
+     5. Iterate until the changes to the direct beam function become small
+
+    Parameters
+    ----------
+    pipelines:
+        List of two pipelines, one for the full wavelength range and one for the bands.
+        The order in which the pipelines are given does not matter.
```
SimonHeybrock, line 64: This is an odd interface.

src/esssans/direct_beam.py
```diff
@@ -0,0 +1,130 @@
...
+        The number of iterations to perform.
+    """
+    if len(pipelines) != 2:
+        raise ValueError("Expected two pipelines, got {}".format(len(pipelines)))
+
+    # Determine which pipeline is for the full wavelength range
+    if pipelines[0].compute(WavelengthBands).ndim == 1:
+        pipeline_full, pipeline_bands = pipelines
+    else:
+        pipeline_bands, pipeline_full = pipelines
```
SimonHeybrock, line 77: Could we simply take one pipeline as input, copy it and set wavelength bands in one of them? The mechanism here is quite unusual, I think.

src/esssans/direct_beam.py
```diff
@@ -0,0 +1,130 @@
...
+        coords={
+            'wavelength': sc.midpoints(wavelength_bands, dim='wavelength')
+            .squeeze()
+            .rename_dims(band='wavelength')
+        },
+    )
+
+    # TODO: This feels like an ugly hack. What is the best way to achieve this?
+    pipeline_full._providers[DirectBeam] = lambda: direct_beam_function
+    pipeline_bands._providers[DirectBeam] = lambda: direct_beam_function
```
SimonHeybrock, line 97: You are bypassing Sciline's check for type annotations?

Why are you not setting this as a param?

src/esssans/direct_beam.py
```diff
@@ -0,0 +1,130 @@
...
+    The steps are as follows:
+
+     1. Create a flat direct beam function, as a function of wavelength, with
+        wavelength bins corresponding to the wavelength bands
+     2. Calculate inside each band by how much one would have to multiply the final
+        $I(Q)$ so that the curve would overlap with the full range curve
+     3. Multiply the direct beam values inside each wavelength band by this factor
+     4. Compare the full-range $I(Q)$ to a theoretical reference and add the
+        corresponding additional scaling to the direct beam function
+     5. Iterate until the changes to the direct beam function become small
```
SimonHeybrock, line 58: But below there is a fixed number of iterations?

src/esssans/beam_center_finder.py
```diff
@@ -381,6 +404,15 @@ def beam_center_from_iofq(
...
 
+    # Flatten positions dim which is required during the iterations for slicing with a
+    # boolean mask
+    pos_dims = data.coords['position'].dims
+    new_dim = uuid.uuid4().hex
+    data = data.flatten(dims=pos_dims, to=new_dim)
+    dims_to_flatten = [dim for dim in norm.dims if dim in pos_dims]
+    if dims_to_flatten:
+        norm = norm.flatten(dims=dims_to_flatten, to=new_dim)
+
```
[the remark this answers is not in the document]
nvaytet: If you are refering to the flattening inside `beam_center_from_center_of_mass`, then yes flattening is still required here because when selecting quadrants defined by a boolean condition on phi, we need the data to be 1d.
You want to flatten here because you want to do it only once, not at every iteration (it would have made more sense to keep the flattening closer to where it is actually needed, i.e. inside `_iofq_in_quadrants`, but it would then be more expensive).

Now you could ask that if we flatten here, maybe we don't need to flatten above? However, we also call `beam_center_from_center_of_mass` directly, without going via `beam_center_from_iofq`, so we also need to flatten there. It just means that if we go via `beam_center_from_iofq`, the flattening in `beam_center_from_center_of_mass` won't have any effect.

src/esssans/i_of_q.py
```diff
@@ -177,36 +195,54 @@ def _to_q_bins(q_bins: Union[int, sc.Variable]) -> Dict[str, Union[int, sc.Varia
...
+    # than to keep all bands in a list and then concatenate the results.
+    band_dim = (set(wavelength_bands.dims) - {'wavelength'}).pop()
+    out = None
+    for wav_range in sc.collapse(wavelength_bands, keep='wavelength').values():
+        band = q_binned.bin(wavelength=wav_range).squeeze()
+        if out is None:
+            out = band
+        else:
+            out = sc.concat([out, band], band_dim)
+    return out
```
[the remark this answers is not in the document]
nvaytet: Any good ideas on how we could do this differently?

src/esssans/i_of_q.py
```diff
@@ -151,17 +158,28 @@ def merge_spectra(
     :
         The input data converted to Q and then summed over all detector pixels.
     """
+    if final_dims is None:
+        final_dims = ['Q']
+    if (wavelength_bands is not None) and (wavelength_bands.ndim == 1):
+        # We expect wavelength_bands to be two-dimensional below
+        wavelength_bands = wavelength_bands.fold(
+            dim='wavelength', sizes={**{'band': 1}, **wavelength_bands.sizes}
+        )
```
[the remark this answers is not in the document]
nvaytet: Yes, I wanted to be able to define the bands via a single variable.

The variable can be 2d (e.g. 10 bands, with a start and end wavelength for each band, so the shape would be `{'band: 10, 'wavelength': 2}`), which allows for overlapping bands.

For convenience, the variable could also be just 1d, a single array of non-overlapping bands (which is a common use case). In that case, I would automatically fold it to a 2d array here,

Reading what I just wrote above,
1. I think this piece of code is wrong, it doesn't do what I want it to do, just adds a dummy dim when it should be making something that has shape `{'band: wavelength_bands.sizes, 'wavelength': 2}`.
2. Because I had to explain all this, I either need to come up with a better interface, e.g. ask for 2 arrays with start and end wavelengths (in that case, there would be a lot of duplicates in the case of non-overlapping bands), or at the very least document it much better in the docstrings.

src/esssans/i_of_q.py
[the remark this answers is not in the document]
nvaytet: Refactored from the comment above to use a `_process_wavelength_bands` function that performs some checks and some optional reshaping.

src/esssans/i_of_q.py
```diff
@@ -177,36 +195,54 @@ def _to_q_bins(q_bins: Union[int, sc.Variable]) -> Dict[str, Union[int, sc.Varia
...
+    sum_dims = set(data_q.dims) - set(final_dims)
     edges = _to_q_bins(q_bins)
     if wavelength_bands is None:
         return data_q.hist(**edges).sum(sum_dims)
-    for i in range(wavelength_bands.sizes['wavelength'] - 1):
-        band = data_q['wavelength', wavelength_bands[i] : wavelength_bands[i + 1]]
+    bands = []
+    band_dim = (set(wavelength_bands.dims) - {'wavelength'}).pop()
+    for wav_range in sc.collapse(wavelength_bands, keep='wavelength').values():
+        band = data_q['wavelength', wav_range[0] : wav_range[1]]
```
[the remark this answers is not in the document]
nvaytet: Refactored from the comment above to use a `_process_wavelength_bands` function that performs some checks and some optional reshaping.

src/esssans/i_of_q.py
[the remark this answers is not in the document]
SimonHeybrock: See also my comments on performance. I think unless overlapping bands are very common we may want to handle this differently anyway, to avoid the performance issues?

src/esssans/i_of_q.py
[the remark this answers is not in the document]
SimonHeybrock: It would probably be faster if we first apply some binning in wavelength, which will allow for pre-selecting bins. But that is already an optimization, I'd rather first think about avoiding this.

Unless bands are overlapping, can we revert to the old approach?

Which gives me an idea:
1. Flatten and sort all the wavelength edges (potentially overlapping).
2. `bin`
3. Assemble bands by combining the (probably up to 3) pieces for each band.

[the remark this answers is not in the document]
SimonHeybrock: Something like this:
```python
def joint(da: sc.DataArray, bands: sc.Variable) -> sc.DataArray:
    binned = da.bin(y=sc.sort(bands.flatten(to='y'), 'y'))
    # At this point we kind of already have what we need, would be cheapest to just return,
    # if follow up providers can work with the result. Otherwise we need to duplicate events:
    sections = []
    for bounds in sc.collapse(bands, keep='y').values():
        # The extra concat can probably be avoided if we insert some dummy edges for first and last band
        sections.append(binned['y',bounds[0]:bounds[1]].bins.concat('y'))
    out = sc.concat(sections, 'band')
    out.coords['y'] = bands
    return out
```

src/esssans/loki/general.py
nvaytet: No, I didn't think of that...
Would you then have a `DetectorBank` parameter like the `RunType` and have `DataWithLogicalDims[DetectorBank, RunType]` everywhere?
Or should we just simply handle this in the loader, inserting required info in the data groups such as bank logical shape and pixel shape?

I guess merging all the contributions to `Q` at the end would also need to merge all the banks...

src/esssans/loki/masking.py
nvaytet: Yes... I think I need to come up with a better way of defining my masks.

src/esssans/loki/masking.py
nvaytet: This relies on something that you pointed out further down which relies on subsequently removing the masks that were used to find the beam center and then adding some new masks back in.

I am starting to doubt the strategy of having some masking before beam center finding, and some other masking after.
I did this because masking the edges of the panel was much easier programmatically after beam center was applied (just an `abs(x)` or `abs(y)`.
But I think I want to go back to the way it was in `main`, with only masking once before beam center finding?

src/esssans/loki/masking.py
nvaytet: See comment above.

src/esssans/direct_beam.py
nvaytet: Yes, let's do that.

src/esssans/loki/io.py
```diff
@@ -0,0 +1,155 @@
...
+    out.coords['source_position'] = source_position
+    out.coords['gravity'] = gravity_vector()
+    return out
+
+
+def _convert_to_tof(da: sc.DataArray) -> sc.DataArray:
+    da.bins.coords['tof'] = da.bins.coords.pop('event_time_offset')
+    if 'event_time_zero' in da.dims:
+        da = da.bins.concat('event_time_zero')
+    return da.bin(tof=1)
```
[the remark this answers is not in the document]
nvaytet: Yes, I removed it and after a few fixed it just works fine :+1:

src/esssans/direct_beam.py
nvaytet: Updated the docstring.

src/esssans/direct_beam.py
nvaytet: What is the best way to make a copy of a pipeline?

```Py
new_pipeline = Pipeline(providers=old_pipeline._providers.values())
for table in old_pipeline._param_tables.values():
    new_pipeline.set_param_table(table)
```
?

src/esssans/direct_beam.py
nvaytet: I think it was because the pipeline is already built, and I found no way to add or replace an existing param.
I guess using the `.insert()` method is probably the way to go.

src/esssans/direct_beam.py
nvaytet: Nope, that only works for actual providers but not with the providers that were created from parameters...

nvaytet: My second attempt in https://github.com/scipp/sciline/pull/91

src/esssans/loki/general.py
nvaytet: Right now, I'm thinking of using a parameter table for the different banks, and then merging all the contributions to $I(Q)$ just before the normalization.

nvaytet: Conclusion is: we leave it as is for now, and we will think about handling multiple bands later.

src/esssans/loki/io.py
```diff
@@ -0,0 +1,155 @@
...
+    )
+
+    if sample_name is None:
+        sample_position = sc.vector(value=[0, 0, 0], unit='m')
+    else:
+        sample_position = out[instrument_path][sample_name]['position']
+    source_position = out[instrument_path][source_name]['position']
+
+    for name in data_entries:
+        out[instrument_path][name][f'{name}_events'] = _preprocess_data(
```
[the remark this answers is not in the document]
nvaytet: Yes, you are correct. I changed the order of things.

src/esssans/i_of_q.py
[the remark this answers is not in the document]
nvaytet: > # The extra concat can probably be avoided if we insert some dummy edges for first and last band

This would only work if we know how many edges to insert, as there is no limit on how wide the bands can be; they could overlap by more than one bin?

src/esssans/beam_center_finder.py
```diff
@@ -68,10 +86,16 @@ def beam_center_from_center_of_mass(
         The beam center position as a vector.
     """
 
-    summed = data.sum(list(set(data.dims) - set(data.coords['position'].dims)))
+    # TODO: the flattening is required for the selection with a boolean mask to work.
+    # Until slicing with a multi-dimensional mask is supported in Scipp, we could avoid
+    # the copy made here by working with the underlying numpy arrays.
```
SimonHeybrock, line 91: It is not going to be supported in Scipp, since it is conceptually impossible (without automatic flattening, which would defeat the purpose of Scipp).

src/esssans/types.py
```diff
@@ -46,6 +53,17 @@ class TransmissionRun(sciline.Scope[RunType, int], int):
...
+NexusInstrumentPath = NewType('NexusInstrumentPath', str)
+
+NexusSampleName = NewType('NexusSampleName', str)
+
+NexusSourceName = NewType('NexusSourceName', str)
+
+NexusDetectorName = NewType('NexusDetectorName', str)
+
+TransformationPath = NewType('TransformationPath', str)
+
```
SimonHeybrock, line 66: Missing docstrings.

src/esssans/types.py
```diff
@@ -82,6 +113,9 @@ class TransmissionRun(sciline.Scope[RunType, int], int):
 CorrectForGravity = NewType('CorrectForGravity', bool)
 """Whether to correct for gravity when computing wavelength and Q."""
 
+FinalDims = NewType('FinalDims', Sequence[str])
+"""Final dimensions of IofQ"""
```
SimonHeybrock, line 117: This needs more detail, it is completely unclear what valid options are, and what the point is.

src/esssans/loki/data.py
```diff
@@ -0,0 +1,33 @@
...
+        # Sample run 2 (sample + sample holder/can, no transmission monitor in beam)
+        '60339-2022-02-28_2215.nxs': 'md5:03c86f6389566326bb0cbbd80b8f8c4f',
+        # Background transmission run (sample holder/can + transmission monitor)
+        '60392-2022-02-28_2215.nxs': 'md5:9ecc1a9a2c05a880144afb299fc11042',
+        # Background run 2 (no sample, sample holder/can only, no transmission monitor)
+        '60393-2022-02-28_2215.nxs': 'md5:bf550d0ba29931f11b7450144f658652',
+        # Sample transmission run (sample + sample holder/can + transmission monitor)
+        '60394-2022-02-28_2215.nxs': 'md5:c40f38a62337d86957af925296c4c615',
+        # Analytical model for the I(Q) of the Poly-Gauss sample
+        'PolyGauss_I0-50_Rg-60.txt': 'md5:389ee172a139c06d062c26b5340ae9ce',
```
SimonHeybrock, line 25: 👍

src/esssans/loki/general.py
```diff
@@ -0,0 +1,58 @@
...
+    NexusInstrumentPath,
+    NeXusMonitorName,
+    NexusSourceName,
+    RunType,
+    TransformationPath,
+    Transmission,
+)
+
+default_parameters = {
+    NexusInstrumentPath: 'instrument',
```
SimonHeybrock, line 23: I wouldn't expect this to change much... or we could change the loader to normalize it? Did you have any concrete need for being able to customize this?

src/esssans/loki/general.py
```diff
@@ -0,0 +1,58 @@
...
+    NexusDetectorName: 'larmor_detector',
+    NeXusMonitorName[Incident]: 'monitor_1',
+    NeXusMonitorName[Transmission]: 'monitor_2',
+    NexusSourceName: 'source',
```
SimonHeybrock, line 27: Some camel-case inconsistencies. One could consider moving the nexus types into a submodule, so we can have `nexus.MonitorName`, `nexus.DetectorName`, etc., but I don't mind too much for now.

src/esssans/loki/io.py
```diff
@@ -0,0 +1,168 @@
...
+    RawMonitor,
+    RunType,
+    TransformationPath,
+    Transmission,
+)
+from .general import default_parameters as params
+
+DETECTOR_BANK_RESHAPING = {
+    params[NexusDetectorName]: lambda x: x.fold(
+        dim='detector_number', sizes=dict(layer=4, tube=32, straw=7, pixel=512)
```
SimonHeybrock, line 35: Why is the *detector name* a function that folds? Also, doesn't this modify `default_parameters`, which looks like it might break things?

Have you considered using a `-1` somewhere, to be more flexible in terms of future changes?
```suggestion
        dim='detector_number', sizes=dict(layer=4, tube=-1, straw=7, pixel=512)
```

src/esssans/loki/masking.py
```diff
@@ -0,0 +1,91 @@
...
+"""Detector tube edge mask"""
+
+
+def detector_straw_mask(
+    sample_straws: RawData[SampleRun],
+) -> DetectorLowCountsStrawMask:
+    return DetectorLowCountsStrawMask(
+        sample_straws.sum('pixel').data
+        < sc.scalar(sample_straws.sizes['pixel'] * 0.5, unit='counts')
+    )
```
SimonHeybrock, line 34: Still worried about such masks, we should never simply mask zero or low count pixels!

src/esssans/loki/masking.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def detector_tube_edge_mask(
+    sample_straws: RawData[SampleRun],
+) -> DetectorTubeEdgeMask:
+    other_dims = set(sample_straws.dims) - {'pixel'}
+    size = np.prod([sample_straws.sizes[dim] for dim in other_dims])
+    return DetectorTubeEdgeMask(
+        sample_straws.sum(other_dims).data < sc.scalar(size * 0.5, unit='counts')
+    )
```
SimonHeybrock, line 54: This does not look like a tube edge mask, but is masking based on counts. Why? I thought we should not masks low counts especially in SANS where low counts at large Q is expected.

tests/loki/loki_directbeam_test.py
SimonHeybrock, line 1: Prefixing everything with the instrument does not seem necessary, since it is in subfolders?

tests/loki/loki_directbeam_test.py
```diff
@@ -0,0 +1,72 @@
...
+    data = np.loadtxt(get_path('PolyGauss_I0-50_Rg-60.txt'))
+    qcoord = sc.array(dims=["Q"], values=data[:, 0], unit='1/angstrom')
+    theory = sc.DataArray(
+        data=sc.array(dims=["Q"], values=data[:, 1], unit=''), coords={"Q": qcoord}
+    )
+    ind = np.argmax((qcoord > q_loc).values)
+    I0 = (theory.data[ind] - theory.data[ind - 1]) / (qcoord[ind] - qcoord[ind - 1]) * (
+        q_loc - qcoord[ind - 1]
+    ) + theory.data[ind - 1]
+    return I0
```
SimonHeybrock, line 26: Same (or very similar) code is found in the docs notebook. Should this become part of the package?

tests/loki/loki_iofq_test.py
```diff
@@ -0,0 +1,71 @@
...
+    params = make_params()
+    band = np.linspace(1.0, 13.0, num=11)
+    params[WavelengthBands] = sc.array(
+        dims=['band', 'wavelength'],
+        values=np.vstack([band[:-1], band[1:]]).T,
+        unit='angstrom',
+    )
+    pipeline = sciline.Pipeline(loki_providers(), params=params)
+    result = pipeline.compute(BackgroundSubtractedIofQ)
+    assert result.dims == ('band', 'Q')
```
SimonHeybrock, line 51: `FinalDims` says `('Q',)`, so this is a bit odd. Is the parameter badly named? Should we instead list which dims not to reduce, or something like that?

tests/normalization_test.py
SimonHeybrock, line 1: I don't think this code should have been moved. While it uses a concrete Sans2d result, it is a test of the generic solid angle code.

src/esssans/loki/io.py
nvaytet: > Why is the detector name a function that folds? Also, doesn't this modify default_parameters, which looks like it might break things?

It's a `:` not a `=`, so I'm just using the name of the detector as a key in a new dict. Isn't that ok?

> Have you considered using a -1 somewhere, to be more flexible in terms of future changes?

I think it would be ok to have that at a later stage. Right now, knowing how inconsistent the files we get may be, I'd rather know immediately if something has changed, instead of this silently reshaping to a different size?

src/esssans/loki/masking.py
nvaytet: We have such masks in [SANS2D](https://github.com/scipp/esssans/blob/main/src/esssans/sans2d.py#L93), so you would also want to change it there?

I agree that masking individual pixels based on counts is not good, for the reasons you stated.
However, the mask here is masking straws that have very few counts when summed over the pixel dimension, which seems to work well for finding dead straws. I have checked the results using different visualizations.

I am open to suggestions as to how we could achieve this otherwise, but right now the alternative I see is to have hard-coded numbers for the y positions of the pixels (e.g. `if y > 0.55 * meter or y < 0.45 * meter`), which would have to change for each run if the beam position has changed?

tests/loki/loki_directbeam_test.py
nvaytet: Should I build a small pipeline that is independent of the $I(Q)$ pipeline which would compute the direct beam function? It would have a provider for I0, and the filename as a parameter?

The provider for the `DirectBeamFunction` would have the iterations loop inside it, instead of just calling a pure python function as it is now.

tests/loki/loki_iofq_test.py
nvaytet: > Should we instead list which dims not to reduce, or something like that?

Yeah, I came up with that `FinalDims` params at some point but never really liked it.
It was a way to be flexible, in case we wanted to return an `I(Q)` per straw or anything we want.
However, that use case is a bit unlikely? We would mostly just want it per layer.

We could either do what you suggest, or we could just have a boolean param that is `perLayer = True/False`?

src/esssans/loki/io.py
SimonHeybrock: > > Why is the detector name a function that folds? Also, doesn't this modify default_parameters, which looks like it might break things?
> 
> It's a `:` not a `=`, so I'm just using the name of the detector as a key in a new dict. Isn't that ok?

Right, I misread.

> 
> > Have you considered using a -1 somewhere, to be more flexible in terms of future changes?
> 
> I think it would be ok to have that at a later stage. Right now, knowing how inconsistent the files we get may be, I'd rather know immediately if something has changed, instead of this silently reshaping to a different size?

src/esssans/loki/masking.py
SimonHeybrock: > We have such masks in [SANS2D](https://github.com/scipp/esssans/blob/main/src/esssans/sans2d.py#L93), so you would also want to change it there?

That is slightly different because it combines a spatial condition with a counts condition. It may still be questionable though.


> I agree that masking individual pixels based on counts is not good, for the reasons you stated. However, the mask here is masking straws that have very few counts when summed over the pixel dimension, which seems to work well for finding dead straws. I have checked the results using different visualizations.

Wouldn't that also risk masking straws that are far from the beam center? How did you determine the threshold?

> I am open to suggestions as to how we could achieve this otherwise, but right now the alternative I see is to have hard-coded numbers for the y positions of the pixels (e.g. `if y > 0.55 * meter or y < 0.45 * meter`), which would have to change for each run if the beam position has changed?

Supposedly dead components will be masked in the NeXus file.

src/esssans/loki/masking.py
SimonHeybrock: A comment should be the absolute minimum. Has this been check with the scientists?

tests/loki/loki_directbeam_test.py
SimonHeybrock: Not everything in the package has to use Sciline. It could just be a function (but it could also be written like a provider, if we want to make a pipeline later).

src/esssans/loki/general.py
SimonHeybrock: The sample (env) is outside the instrument. But yes, we can do what you suggest.

nvaytet: In the end I made a hard-coded variable and removed it from the default parameters.

src/esssans/loki/general.py
nvaytet: I changed everything to `NeXus`.

src/esssans/beam_center_finder.py
```diff
@@ -68,10 +86,15 @@ def beam_center_from_center_of_mass(
         The beam center position as a vector.
     """
 
-    summed = data.sum(list(set(data.dims) - set(data.coords['position'].dims)))
+    # TODO: the flattening is required for the selection with a boolean mask to work.
+    # We could avoid the copy made here by working with the underlying numpy arrays.
+    data = data.flatten(dims=data.coords['position'].dims, to=uuid.uuid4().hex)
```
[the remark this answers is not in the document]
SimonHeybrock: I tried `squeeze` before calling the workflow (outside the beam-center-finder), but it still failed somewhere, can't remember where. Will try again.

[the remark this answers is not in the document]
nvaytet: See https://github.com/scipp/esssans/commit/c9acd436d6cf1a02590f52cbc4d662a0c84f950a

src/esssans/beam_center_finder.py
```diff
@@ -86,11 +86,18 @@ def beam_center_from_center_of_mass(
...
-    summed = data.bins.sum()
+    dims_to_sum = set(data.dims) - set(data.coords['position'].dims)
+    if dims_to_sum:
+        summed = data.sum(dims_to_sum)
+    else:
+        summed = data.bins.sum()
+    if summed.ndim > 1:
+        # TODO: the flattening is required for the selection with a boolean mask to
+        # work. We could avoid the copy made here by working with the underlying numpy
+        # arrays.
```
SimonHeybrock, line 97: What copy? Numpy boolean indexing cannot return without making a copy either?

src/esssans/direct_beam.py
```diff
@@ -7,11 +7,41 @@
...
+
+    Returns
+    -------
+    :
+        The interpolated intensity of the I(Q) for the requested Q value.
+
+    """
+    from .loki.data import get_path
+
+    data = np.loadtxt(get_path(filename))
```
SimonHeybrock, line 33: This function will not be usable for anything but test data then? Can we avoid doing path handling here?

src/esssans/beam_center_finder.py
nvaytet: ah yes, I think that's correct. I'll remove the comment.

src/esssans/beam_center_finder.py
nvaytet: Ok, removed comment.

src/esssans/types.py
nvaytet: Changed to `DimsToKeep` and added a better description.

tests/normalization_test.py
nvaytet: Moved it back.

SimonHeybrock (approved):

nvaytet:
This is currently blocked by https://github.com/scipp/scippnexus/issues/188

SimonHeybrock:
> This is currently blocked by [scipp/scippnexus#188](https://github.com/scipp/scippnexus/issues/188)

Can you use the single-threaded dask scheduler for now?

PR #41, opened 2023-12-19
fix: don't assume positions are relative to sample

jokasimr:
@nvaytet pointed out that there was a difference between the result of the IoQ pipeline after #29 was merged.

See difference between result before and after:
Screenshot at 2023-12-19 11-11-02
Screenshot at 2023-12-19 11-10-41

There seems to be a difference in scale ~15x difference.

The difference seems to be due to 2 factors:
  * The `positions` used in the solid angle computations are relative to the source, while I thought they were relative to the sample.
  * The pixel shapes used previously were determined to have the wrong values. This ought to account for a factor ~2.3 lower IoQ in the result.
  
After changing the solid angle provider to treat the positions as relative to source instead of relative to sample we get: 
Screenshot from 2023-12-19 13-57-29
 
 The scale difference compared to the original result seems to be close to 2.3.


tests/normalization_test.py
```diff
@@ -15,9 +15,6 @@
 
 
 def _sans2d_setup_requirements_for_solid_angle(da):
```
SimonHeybrock, line 17: `da` unused?

src/esssans/normalization.py
```diff
@@ -70,7 +70,7 @@ def solid_angle(
     length = sc.norm(cylinder_axis)
 
     omega = _approximate_solid_angle_for_cylinder_shaped_pixel_of_detector(
-        pixel_position=data.coords['position'],
+        pixel_position=data.coords['position'] - data.coords['sample_position'],
```
SimonHeybrock, line 73: Do we have a unit test with a sample that is not at the origin?

nvaytet:
> The pixel shapes used previously were determined to have the wrong values. This ought to account for a factor ~2.3 lower IoQ in the result.

We really need to verify this with Wojciech, because a factor of 2 will make a very big difference to his results.

src/esssans/normalization.py
jokasimr: No in that case I think that test should have failed in this PR

tests/normalization_test.py
jokasimr: Yes it looks weird. I'll fix that.

src/esssans/normalization.py
jokasimr: Yes we do, I was confused. The current test has a sample not at the origin (its at [0, 0, 19.28]).

The issue was that I thought the `pixel_positions` from the dataarray were relative to the sample, and therefore I didn't subtract the `sample_position` in the provider.
Therefore the tests were made from that assumption, and the data was prepared accordingly, and that's why the tests did not fail before.

SimonHeybrock: Where did the data change in this PR?

SimonHeybrock: Ah, I see, you computed the difference in the test setup?

SimonHeybrock (approved):

src/esssans/normalization.py
jokasimr: Yes exactly.

PR #50, opened 2024-01-16
Zoom data reduction

SimonHeybrock:
This adds a data reduction workflow for Zoom. The main addition is handling of file I/O and reading masks from a series of XML files.

- `esssans.isis.mantidio` provides I/O wrapper depending on Mantid.
- For the docs we are using pooch as usual, bypassing Mantid.
- I found a way of hiding the Pooch `get_path` stuff from the docs notebooks, but adding a provider that converts filenames into paths. This can either be configured to use a configured data folder, or pooch, or anything else someone might come up with (such as search known locations, ...).


src/esssans/isis/__init__.py
```diff
@@ -0,0 +1,43 @@
+# SPDX-License-Identifier: BSD-3-Clause
```
nvaytet, line 1: Should the `sans2d` submodule be part of `isis`?

docs/examples/zoom.ipynb
```diff
@@ -0,0 +1,285 @@
...
+   "execution_count": null,
+   "id": "50f11eb2",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "params[NeXusMonitorName[Incident]] = 'monitor3'\n",
+    "params[NeXusMonitorName[Transmission]] = 'monitor5'\n",
+    "\n",
+    "band = sc.linspace('wavelength', 1.75, 16.5, num=2, unit='angstrom')\n",
+    "params[WavelengthBands] = band\n",
```
nvaytet, line 106: Would be good if we could leave the `WavelengthBands` and only have to specify the `WavelengthBins`?

docs/examples/zoom.ipynb
```diff
@@ -0,0 +1,285 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "fdcaec37",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "providers = sans.providers + sans.isis.providers\n",
+    "providers = providers + (\n",
+    "    sans.isis.transmission_from_background_run,\n",
+    "    sans.isis.transmission_from_sample_run,\n",
```
nvaytet, line 137: Why are those not part of the `isis.providers`?

docs/examples/zoom.ipynb
```diff
@@ -0,0 +1,285 @@
...
+    "pipeline.set_param_table(masks)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "951aec10",
+   "metadata": {},
+   "source": [
+    "If Mantid is available, we can use it to load data files.\n",
+    "**You must configure the `DataFolder` below to point to the directory containing the data files.**\n",
```
nvaytet, line 149: `DataFolder` renders strangely in the docs
Screenshot at 2024-01-18 14-15-30

docs/examples/zoom.ipynb
```diff
@@ -0,0 +1,285 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "558327ff",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "da = iofq.compute()\n",
```
nvaytet, line 214: Maybe we should silence the `pooch` warnings everywhere?
Screenshot at 2024-01-18 14-17-56

docs/examples/zoom.ipynb
```diff
@@ -0,0 +1,285 @@
...
+    "parts = (CleanSummedQ[SampleRun, Numerator], CleanSummedQ[SampleRun, Denominator])\n",
+    "iofqs = (IofQ[SampleRun],)\n",
+    "keys = monitors + (MaskedData[SampleRun],) + parts + iofqs\n",
+    "\n",
+    "results = pipeline.compute(keys)\n",
+    "\n",
+    "display(sc.plot({str(key): results[key] for key in monitors}, norm='log'))\n",
+    "\n",
+    "display(\n",
+    "    scn.instrument_view(\n",
```
nvaytet, line 246: Be aware that having many instrument views in the docs increases size bloat. For a 2d panel maybe we can make a normal plot?

src/esssans/isis/data.py
```diff
@@ -0,0 +1,66 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
```
nvaytet, line 2:
```suggestion
# Copyright (c) 2024 Scipp contributors (https://github.com/scipp)
```

src/esssans/isis/data.py
```diff
@@ -0,0 +1,66 @@
...
+        # ISIS user file configuring the data reduction
+        'USER_ZOOM_Cabral_4m_TJump_233G_8x8mm_Small_BEAMSTOP_v1_M5.toml': 'md5:4423ecb7d924c79711aba5b0a30a23e7',  # noqa: E501
+        # Mask files
+        'andru_test.xml': 'md5:c59e0c4a80640a387df7beca4857e66f',
+        'left_beg_18_2.xml': 'md5:5b24a8954d4d8a291f59f5392cd61681',
+        'right_beg_18_2.xml': 'md5:fae95a5056e5f5ba4996c8dff83ec109',
+        'small_bs_232.xml': 'md5:6d67dea9208193c9f0753ffcbb50ed83',
+        'small_BS_31032023.xml': 'md5:3c644e8c75105809ab521773f9c0c85b',
+        'tube_1120_bottom.xml': 'md5:fe577bf73c16bf5ac909516fa67360e9',
+        'tubes_beg_18_2.xml': 'md5:2debde8d82c383cc3d592ea000552300',
```
nvaytet, line 35: Do you know what the different masks represent? (i.e. what they are trying to mask?)

src/esssans/isis/io.py
```diff
@@ -0,0 +1,76 @@
...
+PixelMaskFilename = NewType('PixelMaskFilename', str)
+CalibrationFilename = NewType('CalibrationFilename', str)
+
+FilenameType = TypeVar('FilenameType', bound=str)
+
+
+DataFolder = NewType('DataFolder', str)
+
+
+class Path(sciline.Scope[FilenameType, str], str):
```
nvaytet, line 22: `Path` will very commonly refer to `pathlib.Path`. Can we avoid using the same name (especially as we sometimes `from types import *`?)

src/esssans/isis/io.py
```diff
@@ -0,0 +1,76 @@
...
+
+MaskedDetectorIDs = NewType('MaskedDetectorIDs', sc.Variable)
+"""1-D variable listing all masked detector IDs."""
+
+
+def to_path(filename: FilenameType, path: DataFolder) -> Path[FilenameType]:
+    return f'{path}/{filename}'
+
+
+def read_xml_detector_masking(filename: Path[PixelMaskFilename]) -> MaskedDetectorIDs:
```
nvaytet, line 38: Is the format of those xml mask files specific to Mantid or are they also used by other software? If the former, should this move into `mantidio`?

src/esssans/isis/mantidio.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def load_run(filename: Path[Filename[RunType]]) -> DataWorkspace[RunType]:
+    loaded = Load(Filename=str(filename), LoadMonitors=True, StoreInADS=False)
+    if isinstance(loaded, Workspace):
+        # A single workspace
+        data_ws = loaded
+    else:
+        # Separate data and monitor workspaces
+        data_ws = loaded.OutputWorkspace
```
nvaytet, line 87: How do we get hold of the monitor data in this case?

src/esssans/isis/__init__.py
SimonHeybrock: Probably, but I want to keep any unrelated changes out of this PR. We can perform such restructuring and rollout of changes in a follow-up.

docs/examples/zoom.ipynb
SimonHeybrock: I copied this from existing notebooks, not sure what you are suggesting.

docs/examples/zoom.ipynb
SimonHeybrock: Because someone might want to set a separate transmission run, and in that case this should not be added.

src/esssans/isis/data.py
SimonHeybrock: Not in detail. Some tubes, beamstop, sample holder, but there appear to be multiple varieties. Why?

src/esssans/isis/io.py
SimonHeybrock: Well, the XML is used elsewhere, but the files are afaik pecific to Mantid.

src/esssans/isis/mantidio.py
SimonHeybrock: Mantid stores a link to the monitor workspace in the data workspace and that is where ScippNeutron pulls the monitors from.

docs/examples/zoom.ipynb
SimonHeybrock: People want to plot calibrated positions though. Is there an easy way of making a 2-D plot with actual positions using the X and Y components?

src/esssans/isis/io.py
SimonHeybrock: Renamed to `FilePath`.

> especially as we sometimes `from types import *`?

I want to get away from that. As you can see I have kept providers and types more structured here, no placing anything in a "meaningless" `types.py`. Still not there yet, but I hope it will work nice eventually.

src/esssans/isis/data.py
nvaytet: I was thinking since we started adding short comments describing what the different files are, we could add some descriptions here?

docs/examples/zoom.ipynb
nvaytet: Just that seeing it again here makes me think we could improve the usability a little more. I think it's a bug, and it's unrelated to this PR, so let's not do anything about it here.

nvaytet: See #51

docs/examples/zoom.ipynb
nvaytet: I guess it would also be a job for https://github.com/scipp/plopp/issues/264 ?

SimonHeybrock: The histogramming solution applies the masks, which is not great. I think we are lucky in this case since `Y` appears to be constant along tubes, so I can fold instead of histogramming.

SimonHeybrock: See update.

src/esssans/isis/data.py
SimonHeybrock: I extended the description.

src/esssans/isis/visualization.py
```diff
@@ -0,0 +1,43 @@
...
+# Copyright (c) 2024 Scipp contributors (https://github.com/scipp)
+"""
+Plotting functions for ISIS data.
+"""
+from typing import Any
+
+import scipp as sc
+
+
+def plot_instrument(da: sc.DataArray, pixels_per_tube: int = 512, **kwargs: Any) -> Any:
```
nvaytet, line 11: Are you thinking this is generic enough that it can be used for other isis instruments? I guess it would work for sans2d as well. But then it should also work for other flat panels, not just isis panels?
Should we call it something else that implies it is plotting a flat panel? As it wouldn't work for something else.

src/esssans/isis/visualization.py
```diff
@@ -0,0 +1,43 @@
...
+    Plot a 2-D instrument view of a data array.
+
+    This is an alternative to the `scn.instrument_view` function, avoiding the 3-D
+    rendering of the instrument. The exact X and Y coordinates of the pixels are
+    used for the 2-D plot.
+
+    Parameters
+    ----------
+    da:
+        The data array to plot. Must have a 'position' coord and a single dimension.
```
nvaytet, line 22: Is it not possible to flatten if there are more than one dimension?

src/esssans/isis/visualization.py
```diff
@@ -0,0 +1,43 @@
...
+        da = da.hist()
+    da.coords['x'] = da.coords['position'].fields.x.copy()
+    da.coords['y'] = da.coords['position'].fields.y.copy()
+    folded = da.fold(da.dim, sizes={'y': -1, 'x': pixels_per_tube})
+    y = folded.coords['y']
+    if sc.all(y.min('x') == y.max('x')):
+        folded.coords['y'] = y.min('x')
+    else:
+        raise ValueError(
+            'Cannot plot 2-D instrument view of data array with non-constant '
```
nvaytet, line 38: Do we really have to raise or would it just make a ragged detector panel, like the other plots we have with 2d coords?

nvaytet: Ah maybe it's because the x coord is already 2d?

src/esssans/isis/visualization.py
```diff
@@ -0,0 +1,43 @@
...
+            'y coordinate along tubes. Use scippneutron.instrument_view instead.'
+        )
+    plot_kwargs = dict(aspect='equal', norm='log', figsize=(6, 10))
```
nvaytet, line 41: I think the `aspect='equal'` make sense for this, but I find it a little strange that it has `log` as a default, and also changes the figsize? This size would not work e.g. for the sans2d panel.

src/esssans/isis/visualization.py
SimonHeybrock: Are there any ISIS SANS instruments that do not have tubes that we have to worry about?

src/esssans/isis/visualization.py
SimonHeybrock: Sure, but this is for ISIS data, which has been loaded with Mantid, so why complicate things? It can be extended once/if required?

src/esssans/isis/visualization.py
SimonHeybrock: Yes, Scipp (and plotting, I guess?) do not support multiple 2-D coords. Or is Plopp supporting this?

src/esssans/isis/visualization.py
SimonHeybrock: I had considered doing this, but then it would complicate since we would need to maybe support summing some dims but flattening others, I didn't think it was worth it at this point. After all, this is not an ESS instrument.

src/esssans/isis/visualization.py
SimonHeybrock: I renamed it.

nvaytet (approved):

SimonHeybrock:
There are still some position offsets to handle (from user file), but since I do not know when we will get the information on what to implement I will merge and handle the remaining bits in a follow-up.

PR #53, opened 2024-01-23
Make wavelength bands truly optional

### LINKED ISSUE #51: The `WavelengthBands` parameter seems like it is optional but it currently isn't
Right now, not specifying the `WavelengthBands` parameter makes a buildable pipeline, which also runs without any errors.
However, the result is drastically different from running it with the param specified, even if it is set to being a single band that contains the min and max of the wavelength bins.

With `WavelengthBands`:
Screenshot at 2024-01-22 15-34-40

Without:
Screenshot at 2024-01-22 15-34-19

I think the issue is because when binning the numerator in Q, we [return early](https://github.com/scipp/esssans/blob/main/src/esssans/i_of_q.py#L243) if there are no bands. However this means that the events in the entire wavelength range have been used (the wavelength binning has been ignored here).

We either need to fix to make sure only events in the range of `WavelengthBins` are used, or just remove the early returning and make a default `WavelengthBands` if it is `None`.

nvaytet:
Fixes #51


tests/sans2d/reduction_test.py
```diff
@@ -100,6 +98,19 @@ def test_pipeline_can_compute_background_subtracted_IofQ_in_wavelength_slices():
     assert result.sizes['band'] == 10
 
 
+def test_pipeline_wavelength_bands_is_optional():
+    params = make_params()
+    pipeline = sciline.Pipeline(sans2d_providers(), params=params)
+    noband = pipeline.compute(BackgroundSubtractedIofQ)
+    with pytest.raises(sciline.UnsatisfiedRequirement):
+        pipeline.compute(WavelengthBands)
```
SimonHeybrock, line 106: I opened https://github.com/scipp/sciline/issues/108 for this.

SimonHeybrock (approved):

PR #56, opened 2024-01-26
Use pytest import-mode=importlib

jl-wynen:
Part of https://github.com/scipp/scipp/issues/3344

### DIFF
--- a/tests/loki/directbeam_test.py
+++ b/tests/loki/directbeam_test.py
@@ -1,15 +1,19 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
+import sys
+from pathlib import Path
 from typing import Callable, List
 
 import sciline
 import scipp as sc
-from common import make_params
 from scipp.scipy.interpolate import interp1d
 
 import esssans as sans
 from esssans.loki.data import get_path
 from esssans.types import DimsToKeep, QBins
+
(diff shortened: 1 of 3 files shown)


tests/loki/directbeam_test.py
```diff
@@ -1,16 +1,20 @@
...
 import sciline
 import scipp as sc
-from common import make_params
 from scipp.scipy.interpolate import interp1d
 
 import esssans as sans
 from esssans.loki.data import get_path
 from esssans.types import DimsToKeep, QBins
 
+sys.path.insert(0, str(Path(__file__).resolve().parent))
```
nvaytet, line 15: It's a little annoying that we have to do this everywhere. Is this the only way?

nvaytet (approved):

tests/loki/directbeam_test.py
nvaytet: Could defining `make_params` in `conftest.py` or something be an alternative?
I guess it could become a dumping ground for all sorts of things that are dealing with unrelated tests, so maybe not the cleanest?

jl-wynen: > Could defining make_params in conftest.py or something be an alternative?

No, you still couldn't import it. But you could turn it into a fixture which would then be available everywhere.

jl-wynen: Alternatively, the path can be updated in `loki/conftest.py` to make that modification local to the loki tests (I think that this works anyway...)

PR #58, opened 2024-01-30
fix: dont reread files in sans2d workflow

### LINKED ISSUE #46: Workflow causes files to be loaded twice when transmission runs are not seperate?
See:

https://github.com/scipp/esssans/blob/a21ff03845a0de37a35d222ddc4634b389f54650/docs/examples/sans2d.ipynb#L71-L74

The workflow will, I think, *not* detect that this is the same file and load it twice.

I think we need to do this mapping later, after loading, somehow.

Note how this is bypassed in the Zoom workflow.

jokasimr:
Fixes #46


docs/examples/zoom.ipynb
```diff
@@ -129,8 +129,8 @@
    "source": [
     "providers = sans.providers + sans.isis.providers\n",
     "providers = providers + (\n",
-    "    sans.isis.transmission_from_background_run,\n",
-    "    sans.isis.transmission_from_sample_run,\n",
+    "    sans.common.transmission_from_background_run,\n",
+    "    sans.common.transmission_from_sample_run,\n",
```
SimonHeybrock, line 133: Maybe importing this top-level in `__init__.py` would make sense? `common` is meaningless to a user, but `sans.transmission_from_sample_run` can be understood?

SimonHeybrock (commented):
Looks good, just one small comment:

SimonHeybrock (approved):

PR #59, opened 2024-01-30
Add mechanism for setting sample and detector bank offset in Zoom workflow

SimonHeybrock:
Related to #52 (but not a fix since the end goal is to parse the file that contains this info).

### DIFF
--- a/docs/examples/zoom.ipynb
+++ b/docs/examples/zoom.ipynb
@@ -1,280 +1,280 @@
 {
-   "cells": [
-      {
-         "cell_type": "markdown",
-         "id": "ba4d302c",
-         "metadata": {},
-         "source": [
-            "# Zoom data reduction\n",
-            "\n",
-            "## Introduction\n",
-            "\n",
-            "This notebook is an example of how ESSsans can be used to reduce data from [Zoom at ISIS](https://www.isis.stfc.ac.uk/Pages/Zoom.aspx).\n",
-            "The following description is kept relatively brief, for more context see the rest of the documentation.\n",
-            "In particular the [Sans2d](./sans2d.ipynb) notebook may be useful."
-         ]
-      },
-      {
-         "cell_type": "markdown",
-         "id": "53e2278d",
-         "metadata": {},
-         "source": [
-            "There are a few things that are not yet handled:\n",
-            "\n",
-            "- TOF or wavelength masks\n",
-            "- Position corrections from user file (not automatically, have manual sample and detector bank offsets)"
-         ]
-      },
-      {
-         "cell_type": "markdown",
-         "id": "09234b87",
-         "metadata": {},
-         "source": [
-            "## Setup\n",
-            "\n",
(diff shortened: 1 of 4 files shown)


docs/examples/zoom.ipynb
```diff
@@ -69,6 +69,8 @@
     "    sans.isis.CalibrationFilename: '192tubeCalibration_11-02-2019_r5_10lines.nxs',\n",
     "    sans.isis.Filename[sans.types.SampleRun]: 'ZOOM00034786.nxs',\n",
     "    sans.isis.Filename[sans.types.EmptyBeamRun]: 'ZOOM00034787.nxs',\n",
+    "    sans.isis.SampleOffset: sc.vector([0.0, 0.0, 0.11], unit='m'),\n",
```
nvaytet, line 72: Are these mostly required because of the IDF mechanism in Mantid/ISIS?
Will the situation be better at ESS with the nexus geometry, or are we expecting the geometry in the files to not be updated for every experiment and often be out of date?

src/esssans/isis/components.py
```diff
@@ -0,0 +1,44 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2024 Scipp contributors (https://github.com/scipp)
+from typing import NewType
+
+import sciline
+import scipp as sc
+
+from ..types import RawData, RunType
+
+PixelMask = NewType('PixelMask', sc.Variable)
```
nvaytet, line 10: This is also defined in `masking.py`. Should it be defined only once?

src/esssans/isis/components.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class ConfiguredRawData(sciline.Scope[RunType, sc.DataArray], sc.DataArray):
```
nvaytet, line 13: I'm not sure about the name. Immediately it wasn't very clear what it was meant to represent.
What kind of other configurations are we expecting apart from custom offsets? Should we have a name that is specific to applying offsets?

docs/examples/zoom.ipynb
SimonHeybrock: I think ECDC *wants* to keep files up to date, but if that will be actually true in practice...

src/esssans/isis/components.py
SimonHeybrock: Do you have a suggestion?

src/esssans/isis/components.py
SimonHeybrock, line 10:
```suggestion
```

src/esssans/isis/components.py
nvaytet: `RawDataWithComponentUserOffsets`?

nvaytet: Should we also rename the provider? I guess `configure_raw_data` made more sense when the output was called `ConfiguredRawData`?

nvaytet (approved):

PR #63, opened 2024-01-31
Faster direct beam

SimonHeybrock:
Might replace #57. Times are now down to 15 second for the normal iteration (used to be 1-2 minutes).

### DIFF
--- a/src/esssans/i_of_q.py
+++ b/src/esssans/i_of_q.py
@@ -265,39 +265,6 @@ def _events_merge_spectra(
 
 
 @timed
-def _as_events(dense: sc.DataArray, dim: str) -> sc.DataArray:
-    """
-    Convert a DataArray containing dense data to event data.
-    """
-    if dim != dense.dims[-1]:
-        raise ValueError(
-            f'Expected dim to be the last dimension, got {dim} '
-            f'while dense data has dimensions {dense.dims}.'
-        )
-    events = sc.DataArray(dense.data.flatten(to='event'))
-    for name, coord in dense.coords.items():
-        if coord.dims == dense.dims:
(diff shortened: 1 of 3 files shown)


src/esssans/i_of_q.py
```diff
@@ -276,9 +278,38 @@ def _dense_merge_spectra(
...
+        ):
+            del stripped.coords[name]
+    to_flatten = [dim for dim in data_q.dims if dim in dims_to_reduce]
+
+    dummy_dim = str(uuid4())
+    flat = stripped.flatten(dims=to_flatten, to=dummy_dim)
+
+    # Apply masks once, to avoid repeated work when iterating over bands
+    mask = irreducible_mask(flat, dummy_dim)
+    # When not all dims are reduce there may be extra dims in the mask and it is not
```
nvaytet, line 298:
```suggestion
    # When not all dims are reduced there may be extra dims in the mask and it is not
```

nvaytet (approved):
Works like a charm :+1:

nvaytet:
Can we also incorporate this change at the same time which keeps all the logical dimensions?
https://github.com/scipp/esssans/pull/57/files#diff-4deee399edfbe1368eec3c758e019a47f2972d9ebecb15878be74a16451f6234

SimonHeybrock:
> Can we also incorporate this change at the same time which keeps all the logical dimensions? https://github.com/scipp/esssans/pull/57/files#diff-4deee399edfbe1368eec3c758e019a47f2972d9ebecb15878be74a16451f6234

Not sure which part your are linking to, just the removal of the flattening?

nvaytet:
> Not sure which part your are linking to, just the removal of the flattening?

Yes, basically to keep the dimensions as `layer=4, tube=32, straw=7, pixel=512`.

PR #66, opened 2024-02-01
Do not flatten dims when loading loki detector bank

nvaytet:
This keeps all logical dimensions in the loaded loki data, so slicing with `data['layer', 0]['straw', 1]` would give a single flat layer of pixels.

At the same time, we make sure we can compute an $I(Q)$ or a direct beam function per layer and straw.


SimonHeybrock (approved):

SimonHeybrock (approved):

src/esssans/i_of_q.py
```diff
@@ -171,7 +172,7 @@ def merge_spectra(
     data: CleanQ[RunType, IofQPart],
     q_bins: QBins,
     wavelength_bins: WavelengthBins,
```
SimonHeybrock, line 174: @nvaytet Is this unused now? Looks like it.

As I am working on a follow-up change, I can just do it there, if you confirm that it should indeed be removed.

PR #67, opened 2024-02-01
Importable mantidio

### LINKED ISSUE #55: Docs build issues with `isis.mantidio`
Run `tox -e docs` to reproduce:
```
WARNING: [autosummary] failed to import esssans.isis.mantidio.
Possible hints:
* ImportError: 
* AttributeError: module 'esssans.isis' has no attribute 'mantidio'
* ModuleNotFoundError: No module named 'mantid'
```

Should we try to make this importable without Mantid after all?

jl-wynen:
Fixes #55

### DIFF
--- a/src/esssans/isis/mantidio.py
+++ b/src/esssans/isis/mantidio.py
@@ -27,9 +27,9 @@ except ModuleNotFoundError:
     # Catch runtime usages of Mantid
     class _MantidFallback:
         def __getattr__(self, name: str) -> NoReturn:
-            raise RuntimeError(
+            raise ImportError(
                 'Mantid is required to use `sans.isis.mantidio` ' 'but is not installed'
-            )
+            ) from None
 
     _mantid_api = _MantidFallback()
     _mantid_simpleapi = _MantidFallback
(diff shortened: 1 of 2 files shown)


src/esssans/isis/mantidio.py
```diff
@@ -21,6 +19,24 @@
...
 
+try:
+    import mantid.api as _mantid_api
+    import mantid.simpleapi as _mantid_simpleapi
+    from mantid.api import MatrixWorkspace
+except ModuleNotFoundError:
+    # Catch runtime usages of Mantid
+    class _MantidFallback:
+        def __getattr__(self, name: str) -> NoReturn:
+            raise RuntimeError(
```
SimonHeybrock, line 30: Do we need to update the exception type in the Zoom notebook?

jl-wynen: No because it imports mantid explicitly.

But I was actually wondering if it makes sense to raise `ImportError` here. Is there a guideline on this?

SimonHeybrock: Let us go with `ImportError`?

SimonHeybrock (approved):

PR #68, opened 2024-02-05
Refactor to handle wavelength bands during normalization step

SimonHeybrock:
Aside from simplifying code, this has the advantage of offering more flexibility. For example, it will allow for running the direct-beam iteration in the pre-computed `CleanSummedQ`, with minimal cost per iteration, i.e., an overall speedup.

This would also allow for pre-computing a modified denominator term containing no monitors. Monitors could then be factor in in the final step. This would also reduce the impact of variances of monitor broadcasting, since a broadcast to all pixel is avoided (only broadcast to all Q). We can explore this in a follow-up.


src/esssans/i_of_q.py
```diff
@@ -236,87 +217,43 @@ def _to_q_bins(q_bins: Union[int, sc.Variable]) -> Dict[str, Union[int, sc.Varia
...
-    for bounds in sc.collapse(wavelength_bands, keep=dim).values():
-        # The extra concat can probably be avoided if we insert some dummy edges for
-        # first and last band, but we would need to know how many edges to insert, as
-        # the bands can be very wide and overlap by more than one bin.
-        sections.append(wav_binned[dim, bounds[0] : bounds[1]].bins.concat(dim))
-    band_dim = (set(wavelength_bands.dims) - {'wavelength'}).pop()
-    out = sc.concat(sections, band_dim)
-    out.coords[dim] = wavelength_bands
-    return out
+    q_all_pixels = data_q.bins.concat(dims_to_reduce)
```
nvaytet, line 226: Now that the body of this function is so small, and the one for the dense merging is also much reduced, would it make more sense to just inline the code in `merge_spectra`?

I don't mind, it's a matter of taste. It may still be more readable as it is now.

src/esssans/i_of_q.py
```diff
@@ -236,87 +217,43 @@ def _to_q_bins(q_bins: Union[int, sc.Variable]) -> Dict[str, Union[int, sc.Varia
...
-        if name not in ['Q', 'wavelength'] and any(
-            [dim in dims_to_reduce for dim in coord.dims]
-        ):
+        if name not in {'Q', 'wavelength'} and set(coord.dims) & dims_to_reduce:
             del stripped.coords[name]
     to_flatten = [dim for dim in data_q.dims if dim in dims_to_reduce]
 
-    dummy_dim = str(uuid4())
-    # Make sure that dims to flatten are contiguous, and that Q is the last dim
+    # Make dims to flatten contiguous, keep wavelength is the last dim
```
nvaytet, line 247:
```suggestion
    # Make dims to flatten contiguous, keep wavelength as the last dim
```

nvaytet (approved):
Looks good

src/esssans/i_of_q.py
SimonHeybrock: Done

nvaytet (approved):

PR #69, opened 2024-02-05
Very fast direct beam iteration

SimonHeybrock:
Follow up to #63 (which reduced time from 1 minute to 15 seconds). This brings it down to about 5 seconds, essentially the time of a norm reduction in this case.

- The iteration itself is very cheap now, timings are almost independent of the iteration count now.
- Computing by, e.g., layers has almost the same cost as without layers.

This used #68, which allows for computation of the $I(Q)$ numerator and denominator as a function of $(\lambda, Q)$. We can then simply scale this, without re-performing the costly spectrum merge. If addition dimensions are kept then the numerator and denominator will have additional dimensions, making them large (and thus everything slightly slower). Unless we do this for each pixel I do not expect a big impact. If required, may be want to switch back to using events in this case.

As a follow-up, we may consider refactoring the direct-beam interface: Since the cost is now mainly in the "setup", we could turn this into an object that can then be used (with minimal cost after creation) to run iterations, change band config, ...

### DIFF
--- a/src/esssans/direct_beam.py
+++ b/src/esssans/direct_beam.py
@@ -127,9 +127,7 @@ def direct_beam(pipeline: Pipeline, I0: sc.Variable, niter: int = 5) -> List[dic
 
     results = []
 
-    for it in range(niter):
-        print("Iteration", it)
-
+    for _it in range(niter):
         # The first time we compute I(Q), the direct beam function is not in the
         # parameters, nor given by any providers, so it will be considered flat.
         # TODO: Should we have a check that DirectBeam cannot be computed from the
--- a/src/esssans/direct_beam.py
+++ b/src/esssans/direct_beam.py
@@ -152,7 +152,7 @@ def direct_beam(pipeline: Pipeline, I0: sc.Variable, niter: int = 5) -> List[dic
             I0=I0,
         )
 
-        # Scale denominator terms that where initially computed without direct beam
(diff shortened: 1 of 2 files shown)


src/esssans/direct_beam.py
```diff
@@ -99,44 +98,44 @@ def direct_beam(pipeline: Pipeline, I0: sc.Variable, niter: int = 5) -> List[dic
...
         CleanSummedQ[BackgroundRun, Numerator],
+        CleanSummedQ[BackgroundRun, Denominator],
     )
+    parts = pipeline.compute(parts)
+    # Convert events to histograms to make normalization (in every iteration) cheap
+    for key in [
+        CleanSummedQ[SampleRun, Numerator],
+        CleanSummedQ[BackgroundRun, Numerator],
+    ]:
+        parts[key] = parts[key].hist(wavelength=wavelength_bins)
```
nvaytet, line 118: Should we check here that `parts[key].bins is not None`?

nvaytet: Or maybe we leave it because we want an error to be raised if non-event data is being fed through so that we are not silently doing things we don't expect?

SimonHeybrock: Yes, should we leave it until we get a requirement to make this work for histogrammed data?

nvaytet (approved):
Tested it locally :+1:

PR #71, opened 2024-02-06
Add support for computing I(Q_x, Q_y)

### LINKED ISSUE #60: Add option to compute `I(Q_x, Q_y)`
Would use scipp/scippneutron#490.

SimonHeybrock:
Fixes #60.

This is ready for review, apart from addition of unit tests, which I am writing now.


src/esssans/i_of_q.py
```diff
@@ -178,10 +190,21 @@ def merge_spectra(
...
+        if len(edges) == 1:
+            out = flat.hist(**edges)
+        else:
+            # sc.hist (or the underlying sc.bin) cannot deal with extra data dims,
+            # work around by flattening and regrouping.
+            out = (
+                flat.flatten(to=str(uuid.uuid4()))
+                .group(*[flat.coords[dim] for dim in flat.dims if dim != helper_dim])
+                .hist(**edges)
+            )
```
nvaytet, line 207: For my understanding, can you explain how this works for Qxy. `helper_dim` is always `Q`, so I'm not really following what is happening.

src/esssans/normalization.py
```diff
@@ -406,11 +406,7 @@ def _reduce(da: sc.DataArray) -> sc.DataArray:
                 )
     elif numerator.bins is not None:
         numerator = numerator.hist()
-    if numerator.bins is not None and denominator.bins is None:
-        da = numerator.bins / sc.lookup(func=denominator, dim='Q')
-    else:
-        da = numerator / denominator
-    return IofQ[RunType](da)
```
nvaytet, line 413: Not sure I understand why we don't need this anymore?

tests/loki/iofq_test.py
```diff
@@ -102,16 +108,17 @@ def test_pipeline_can_compute_IofQ_in_overlapping_wavelength_bands():
...
-def test_pipeline_can_compute_IofQ_in_layers():
-    params = make_params()
+@pytest.mark.parametrize('qxy', [False, True])
+def test_pipeline_can_compute_IofQ_in_layers(qxy: bool):
+    params = make_params(qxy=qxy)
     params[DimsToKeep] = ['layer']
     pipeline = sciline.Pipeline(loki_providers(), params=params)
     result = pipeline.compute(BackgroundSubtractedIofQ)
-    assert result.dims == ('layer', 'Q')
+    assert result.dims == ('band', 'Qy', 'Qx') if qxy else ('band', 'Q')
```
nvaytet, line 121: I don't understand why `layer` has changed to `band`.

src/esssans/i_of_q.py
SimonHeybrock: The dims that were "kept" have been flattened and regrouped, but we do not want to group in the dim that we will `hist` afterwards.

src/esssans/normalization.py
SimonHeybrock: It was not actually related to this change, but since we, e.g., `bin(Q=...)` in `merge_spectra` we get the same normalization term for all events in a bin. The `lookup` was not actually required (or never), since we always performed such Q binning (or now: Qx, Qy binning).

tests/loki/iofq_test.py
SimonHeybrock: Copy and paste error from above :grimacing:

SimonHeybrock:
I added more tests now, please have another look!

src/esssans/i_of_q.py
SimonHeybrock: I am wondering if we should just convert the denominator to "events". This would unify a bunch of handling in the merge and normalization. I will consider this separately, not in this PR.

nvaytet:
Memory issue again?

SimonHeybrock:
> Memory issue again?

:cry:

src/esssans/i_of_q.py
nvaytet: > I am wondering if we should just convert the denominator to "events"

How do you do that? Make one event per bin and give it the counts in the bin as weights? (And by bin I mean the original binning of the monitor data)

nvaytet (approved):

src/esssans/i_of_q.py
SimonHeybrock: Yes, exactly.

PR #72, opened 2024-02-06
Port beam center finder notebook from old docs and fix bug in center finder

nvaytet:
- Porting the old notebook from https://scipp.github.io/ess/techniques/sans/sans-beam-center-finder.html

- Also fixed a bug in the beam center finder based on I(Q) where `phi` positions for selecting pixels in the 4 quadrants were being computed on the raw uncalibrated data, instead of the calibrated data. This meant that while we were computing I(Q) using the correct pixel positions, when comparing 4 quadrants to get the curves to overlap, we were comparing data from quadrants not symmetrically divided around the newly found center, but around the origin `[0, 0]`.

### DIFF
--- a/docs/examples/beam-center-finder.ipynb
+++ b/docs/examples/beam-center-finder.ipynb
@@ -103,7 +103,7 @@
     "\n",
     "Adding a mask, based on the pixel neutron counts (integrated along the `tof` dimension),\n",
     "will yield a circular mask around the beam center.\n",
-    "Because the sample holder is highly absorbant to neutrons, such a mask will also mask out the sample holder,\n",
+    "Because the sample holder is highly absorbent to neutrons, such a mask will also mask out the sample holder,\n",
     "making it a very simple but effective way of masking."
    ]
   },
--- a/docs/examples/beam-center-finder.ipynb
+++ b/docs/examples/beam-center-finder.ipynb
@@ -10,11 +10,10 @@
     "In SANS experiments, it is essential to find the center of the scattering pattern in order to allow symmetric summation of the scattering intensity around the beam (i.e. computing a one-dimensional $I(Q)$).\n",
     "As detector panels can move, the center of the beam will not always be located at the same place on the detector panel from one experiment to the next.\n",
     "\n",
-    "Here we describe three different algorithms that can be used to determine the position of the beam center:\n",
+    "Here we describe two different algorithms that can be used to determine the position of the beam center:\n",
     "\n",
-    "1. using the center of mass of the pixel counts\n",
-    "1. using an iterative refinement on a computed scattering cross-section to find the center of the scattering pattern\n",
-    "1. a combination of 1. and 2."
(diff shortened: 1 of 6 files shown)


src/esssans/beam_center_finder.py
```diff
@@ -252,13 +256,17 @@ def _cost(xy: List[float], *args) -> float:
...
             'Non-finite value computed in cost. This is likely due to a division by '
-            'zero. Try restricting your Q range, or increasing the size of your Q bins '
-            'to improve statistics in the denominator.'
+            'zero. The value was artificially changed to a large finite number. '
+            'If the results are not satisfactory, try restricting your Q range, or '
+            'increasing the size of your Q bins to improve statistics in the '
+            'denominator.'
         )
+    logger.info(f'Beam center finder: x={xy[0]}, y={xy[1]}, cost={out}')
     return out
```
SimonHeybrock, line 270: Is this an intentional double use of `out`? So this is returning something very different if it is not finite?

src/esssans/beam_center_finder.py
```diff
@@ -158,9 +159,6 @@ def _iofq_in_quadrants(
         'north-west'.
     """
     pi = sc.constants.pi.value
-    phi = data.transform_coords(
-        'phi', graph=graph, keep_intermediate=False, keep_inputs=False
-    ).coords['phi']
     phi_bins = sc.linspace('phi', -pi, pi, 5, unit='rad')
```
SimonHeybrock, line 162: Take care here when resolving the conflict.

src/esssans/beam_center_finder.py
nvaytet: If `out` is NaN, the optimizer will either fail or go round in circles and never finds a converged value.
So I'm returning a large cost instead of a NaN. I don't think I can modify the value anywhere else, because this function is used directly by the scipy wrapper.

Before, we used to raise as soon as the optimizer was trying to probe a region of space where there was almost no data. But actually, returning a large cost allows it to continue and look elsewhere.

Now that I think of it, it may work if I return `np.inf` instead of a large number. That would probably be better.

docs/examples/beam-center-finder.ipynb
```diff
@@ -0,0 +1,559 @@
...
+    "p.ax.plot(xc.value, yc.value, 'o', color='deepskyblue', ms=5)\n",
+    "p"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "35ba00f1-705e-4307-b338-893323bf3e20",
+   "metadata": {},
+   "source": [
+    "## Method 2: computing $I(Q)$ inside 4 quadrants"
```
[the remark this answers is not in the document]
nvaytet: See update

src/esssans/beam_center_finder.py
nvaytet: I hope I didn't mess up

SimonHeybrock (approved):

PR #77, opened 2024-02-09
Merge sans2d and zoom code into isis module

### LINKED ISSUE #48: Gravity vector is needed even if gravity is not used to compute two_theta
In the conversion graph, we are [adding coordinates](https://github.com/scipp/esssans/blob/main/src/esssans/conversions.py#L161-L165) (cylindrical_x, cylindrical_y unit vectors) that are used to compute the `phi` angle which is then used to decompose the detector panel in phi quadrants to [compute the beam center](https://github.com/scipp/esssans/blob/main/src/esssans/beam_center_finder.py#L179).
Those make use of the direction of gravity to determine what is down.

This means that even if we don't want to apply gravity correction to the flight paths of the neutrons, we still need the gravity coordinate.
However, if we are only using the beam center finder based on the center of mass (as opposed to $I(Q)$ inside 4 quadrants), then we do not need the gravity vector at all, but the workflow will complain if the gravity vector is missing from the data.

We should probably improve the logic to only require the gravity vector if needed.

nvaytet:
This unifies the sans2d and zoom code into the `isis` module.

I also used `mantidio` to make new files for sans2d. In the process, I also dropped the dummy `tof` coordinate from the data which just had a single (not very useful) bin. So I also made new files for zoom.

Fixes #48  because the new files all have the gravity coordinate added.
Gravity is needed to know which direction is up/down to apply beam center corrections perpendicular to the beam

Edit: Converting to draft because there is a small difference in the results. We are now just masking the second detector panel, when before it was sliced out. Maybe it can make a small difference to the beam center finder?


src/esssans/loki/general.py
```diff
@@ -51,6 +52,16 @@ def get_monitor_data(
     return RawMonitor[RunType, MonitorType](out)
 
 
+def calibrate_monitor_position(
```
nvaytet, line 55: I don't like the fact that I needed to add this. Should the `loki.get_monitor_data` just return `CalibratedMonitor` instead of `RawMonitor`?

docs/examples/beam-center-finder.ipynb
```diff
@@ -246,8 +246,14 @@
     "params[WavelengthBins] = sc.linspace(\n",
     "    'wavelength', start=2.0, stop=16.0, num=141, unit='angstrom'\n",
     ")\n",
-    "params[FileList[TransmissionRun[SampleRun]]] = params[FileList[SampleRun]]\n",
-    "params[FileList[EmptyBeamRun]] = ['SANS2D00063091.hdf5']\n",
+    "params[isis.Filename[TransmissionRun[SampleRun]]] = params[isis.Filename[SampleRun]]\n",
```
SimonHeybrock, line 249: See #46, it seems it was forgotten to apply the fix here?

src/esssans/common.py
```diff
@@ -63,17 +63,15 @@ def mask_range(
...
-                values=np.union1d(edges.values, da.coords[dim].values),
-                unit=edges.unit,
-            )
-            out = da.bin({dim: new_bins})
-            out.masks[name] = sc.lookup(lu, dim)[sc.midpoints(new_bins, dim=dim)]
-        else:
-            out = da.bin({dim: edges})
-            out.masks[name] = mask.data
+        if dim not in da.coords:
+            da = da.bin({dim: 1})
```
SimonHeybrock, line 67: Why do we have to bin twice? That can be costly.

Edit: I see now it was moved here from `conversions.py`. Still my question stands.

src/esssans/isis/io.py
```diff
@@ -31,10 +31,6 @@ class Filename(sciline.Scope[RunType, str], str):
 """1-D variable listing all masked detector IDs."""
 
 
-def to_path(filename: FilenameType, path: DataFolder) -> FilePath[FilenameType]:
-    return f'{path}/{filename}'
-
-
```
SimonHeybrock, line 37: Why was this removed? This is needed for generic instrument-independent (and pooch-independent) setup. My thought was that this would actually move to the top level eventually, i.e., not remain in the `isis` module.

src/esssans/isis/io.py
```diff
@@ -77,4 +73,12 @@ def read_xml_detector_masking(
...
-providers = (read_xml_detector_masking, to_path)
+def load_run(filename: FilePath[Filename[RunType]]) -> LoadedFileContents[RunType]:
+    return LoadedFileContents[RunType](sc.io.load_hdf5(filename))
+
+
+def load_direct_beam(filename: FilePath[DirectBeamFilename]) -> DirectBeam:
+    return DirectBeam(sc.io.load_hdf5(filename))
+
+
+providers = (read_xml_detector_masking, load_run, load_direct_beam)
```
SimonHeybrock, line 84: I don't feel these belong here, since they only work with some special files that we made for the docs. `isis.io` should remain clear of that, i.e., contain only functionality that works on "regular" files.

src/esssans/isis/zoom.py
SimonHeybrock, line 1: I would avoid naming this `isis.zoom`. It contains special data for docs/testing, not just a general workflow for Zoom. I think we should very clearly separate the two. Maybe `isis.zoom_data`?

src/esssans/loki/general.py
SimonHeybrock: Yes, if we know that the raw files contain calibrated positions then we should directly return that.

nvaytet: I guess it depends what we mean by "calibrated". Does it mean that the positions have been modified to be correct, or does it simply mean that the positions are correct?

If the former, then we can't really say that this would return `CalibratedMonitor` because the positions have not been modified, they are "raw".

src/esssans/common.py
nvaytet: I think it was a very lazy way of getting bin bounds that encompassed all the events.
I now changed it to computing the limits by hand if needed.

src/esssans/isis/zoom.py
nvaytet: Should I then just make 2 sub-folders for `sans2d` and `zoom`?

SimonHeybrock: Why?

nvaytet: Cause I thought it would be a little strange to have a `zoom_data.py`, and a `sans2d.py` file that contains both things to do with (pooch) data and masking. So I wanted to make:
```
sans2d/data.py
sans2d/masking.py
zoom/data.py
```

I guess we could also just make
```
sans2d_data.py
sans2d_masking.py
zoom_data.py
```
but I thought the separation would be clearer with subfolders.

nvaytet: Alternatively, I could make a common `data.py` file for both zoom and sans2d, and then just have a `sans2d.py` file that contains masking providers (where future sans2d-specific providers could then be added).

SimonHeybrock: Why would you have `sans2d` contain both things? We should treat the two in the same manner and split the files by purpose.

SimonHeybrock: I thought masking from masks files is ISIS-specific, not instrument specific, so I would leave that as `isis.masking`?

src/esssans/common.py
```diff
@@ -64,12 +64,19 @@ def mask_range(
...
+                edges.values,
+                np.array(
+                    [
+                        underlying.min().value,
+                        np.nextafter(underlying.max().value, np.inf),
+                    ]
+                ),
+            )
+        else:
+            new_bins = np.union1d(edges.values, da.coords[dim].values)
```
SimonHeybrock, line 78: What if the units do not match?

src/esssans/loki/general.py
nvaytet: I changed to return `CalibratedMonitor` directly.

src/esssans/common.py
nvaytet: I think it's actually ok in this particular case, because below, we do
```Py
new_bins = sc.array(dims=[dim], values=new_bins, unit=edges.unit)
out = da.bin({dim: new_bins})
```
So the bad unit would be caught by the `da.bin` (I think?).

However, it's more a matter of luck than anything else. I will add an explicit check.

nvaytet:
So I've narrowed down the difference in results between the old implementation and new implementation for the SANS2D workflow to the values of the variances.

The data values on the final I(Q) are identical.

For the variances, if I use the default pipeline, I get larger variances (visible ay high Q) in the new implementation.
Screenshot at 2024-02-12 16-42-45

If I use the mode which drops the variances when broadcasting, the curves seem to overlap
Screenshot at 2024-02-12 16-43-37

I think it comes from the fact that the difference between the old and the new is that in the new, we are keeping both detector panels (and masking the second panel), while in the old we were slicing that panel out when loading the data.
I think if we are broadcasting variances, we are broadcasting to the number of pixels, not looking at the masks, so the variances could still end up being larger, even if we have masked the pixels?

Interestingly, looking directly at the variances of the final result (`sc.variances(iofq)`) shows that even with the `UncertaintyBroadcastMode.drop` mode, the variances in the new implementation are still larger that in the old one.
Screenshot at 2024-02-12 16-46-36

Is this because there are still some reduction operations that occur over all the pixels?

src/esssans/isis/io.py
nvaytet: I put this back in when I made a common `data.py` file. See if that works.

src/esssans/common.py
SimonHeybrock: I think we should add a conversion (or check) of the mask edges to the correct unit close to the beginning of the function, where also other checks are performed. Otherwise we get a cryptic error from `sc.bin` here, or worse, someone will modify the code and the coincidental unit check will suddenly disappear.

SimonHeybrock (approved):

SimonHeybrock:
> I think if we are broadcasting variances, we are broadcasting to the number of pixels, not looking at the masks, so the variances could still end up being larger, even if we have masked the pixels?

That may be it. We should probably take into account masks in the upper-bound estimate, i.e., exclude masked pixels from the scale factor? Can you try if you observe the same difference in event-mode?

> Is this because there are still some reduction operations that occur over all the pixels?

Can't think of anything. The solid angle should ignore masking. Are you sure the same number of non-masked pixels is used? Is there some duplication?

nvaytet:
It seems I can no longer reproduce the differences above if I set `params[UncertaintyBroadcastMode] = UncertaintyBroadcastMode.drop`

I set up the params and then did
```Py
providers = (
    sans.providers + isis.providers + isis.data.providers + isis.sans2d.providers
)
providers = list(providers + (
    sans.transmission_from_background_run,
    sans.transmission_from_sample_run,
))

pipeline_new = sciline.Pipeline(providers=providers, params=params)

pipeline_old = sciline.Pipeline(providers=providers, params=params)
pipeline_old.insert(isis.general.get_detector_data_sliced)
```
where
```Py
def get_detector_data_sliced(
    dg: LoadedFileContents[RunType],
) -> RawData[RunType]:
    return RawData[RunType](dg['data']['spectrum', :61440].copy())
```

Then, when computing IofQ
```Py
t = BackgroundSubtractedIofQ
old = pipeline_old.compute(t)
new = pipeline_new.compute(t)
```
both results have identical variances.

I must have messed up in my notebook above.

nvaytet:
> Can you try if you observe the same difference in event-mode?

Results are the same in dense and event modes.

nvaytet:
> We should probably take into account masks in the upper-bound estimate, i.e., exclude masked pixels from the scale factor?

Should we do that in a separate PR?

SimonHeybrock:
> > We should probably take into account masks in the upper-bound estimate, i.e., exclude masked pixels from the scale factor?
> 
> Should we do that in a separate PR?

Yes, but I am not sure what you said above. Is the issue resolved otherwise, or is there still an unknown source of difference?

nvaytet:
> Is the issue resolved otherwise, or is there still an unknown source of difference?

I will check one last time today, but I think there is no difference if we drop the variances. I think I must have messed up in my notebook when I first reported the issue. I will also check with a quick fix that if we take into account the masks in the broadcast of variances, it is also ok.

nvaytet:
I found the bug in my notebook. I was checking `sc.variances(new.hist() - old.hist())` instead of `sc.variances(new.hist()) - sc.variances(old.hist())`. The latter is zero everywhere.

nvaytet:
I have now also verified that taking the masks into account when doing the variance broadcast yields the same results as before.

PR #81, opened 2024-02-12
Optimize spectra merge and normalization

SimonHeybrock:
This adds the following optimizations:

- Bypass some multi-threading shortcomings of DataArray.bin
- Avoid bins.concat where possible
- Use in-place normalization

I have tried this on a 10 GByte file and it shaves off several seconds of runtime when returning event data.


src/esssans/loki/io.py
```diff
@@ -38,6 +38,8 @@ def _patch_data(
 ) -> sc.DataArray:
     out = da.copy(deep=False)
     if out.bins is not None:
+        # Currently ScippNexus adds 32 bit event weights, this may cause trouble.
+        out.bins.data = out.bins.data.to(dtype='float64', copy=False)
```
nvaytet, line 42: Does this also happen with the ISIS files?

SimonHeybrock: I think so actually. We should probably have a chat about this. It would be nice if we could keep 32 bit until needed, so I was not happy about changing this here.

SimonHeybrock: I reverted this change (and fixed the tests instead). Previously there was an (maybe unintentional) conversion done in `normalize`, which this PR remove.

I opened #88 instead.

Can you have another brief look @nvaytet?

tests/loki/iofq_test.py
```diff
@@ -81,13 +81,17 @@ def test_pipeline_can_compute_IofQ_in_event_mode(uncertainties, target, qxy: boo
     assert sc.allclose(
         sc.values(reference.data),
         sc.values(result.hist().data),
-        rtol=sc.scalar(1e-11),
-        atol=sc.scalar(1e-11),
+        # Could be 1e-11, but currently the workflow defaults to float32 data, as
+        # returned by ScippNexus.
+        rtol=sc.scalar(1e-7),
```
nvaytet, line 86: Not sure I follow anymore. So the changes you made here have degraded the accuracy of the results?

SimonHeybrock: Yes, because I change `numerator / denominator` to `numerator /= denominator`. So previously we ended up converting to `float64` in that step.

nvaytet (approved):

PR #91, opened 2024-02-15
Temporary fix for Loki masking

nvaytet:
We had an issue with masking when merging multiple runs in the Loki workflow,

For example, to mask the straws at the top and bottom of the panel, we summed the counts inside each straw, and compared that to the number of pixels in each straw. This means that all you need to do is merge the events from 2 runs, and you will have approximately double the events but the number of pixels stays the same.
So now, in some straws, double the events means you are above the threshold and you start to mask less straws. The more runs you merge, the less straws you mask.

I've 'fixed' this by using a different threshold (based on the max of the counts integrated inside one straw), but this is still based on counts, and will probably break again in the future.

We need to sit down and come up with a better way of masking things.

### DIFF
--- a/src/esssans/loki/masking.py
+++ b/src/esssans/loki/masking.py
@@ -26,25 +26,23 @@ DetectorTubeEdgeMask = NewType('DetectorTubeEdgeMask', sc.Variable)
 
 
 def detector_straw_mask(
-    sample_straws: RawData[SampleRun],
+    data: RawData[SampleRun],
 ) -> DetectorLowCountsStrawMask:
     """
     This mask aims to remove straws with low counts. It is based on the assumption
-    that the integrated counts in a straw should be at least greater than the number of
-    pixels in the straw (within a fudge factor of 0.5).
+    that the integrated counts in a straw should be at least greater than a fraction of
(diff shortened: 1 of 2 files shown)


SimonHeybrock:
Was this suggested by the Scientists? I suppose this yields a sort of rectangular mask, so we include large Q in the four "diagonal" directions?

SimonHeybrock (approved):

nvaytet:
> Was this suggested by the Scientists?

I have asked for Wojciech's opinion, but he's on vacation so we'll talk about it when he returns next week.

> I suppose this yields a sort of rectangular mask, so we include large Q in the four "diagonal" directions?

It yields such a mask, which mimics the rectangular mask that is made in Mantid:
Screenshot at 2024-02-19 10-24-03

We do include large Q in the diagonals yes.
It is not masking the exact same straws as Mantid. We probably want to move to xml masks like we did in Zoom, but this 'fix' is at least equivalent to what we had before, but better if you add multiple runs together.

We could also just get hold of the xml masks if they exist and replace this PR with such masks. I will ask on Slack.

PR #96, opened 2024-02-20
Use pixel masks from XML file for Loki

nvaytet:
Replaces #91 

We had an issue with masking when merging multiple runs in the Loki workflow,

For example, to mask the straws at the top and bottom of the panel, we summed the counts inside each straw, and compared that to the number of pixels in each straw. This means that all you need to do is merge the events from 2 runs, and you will have approximately double the events but the number of pixels stays the same.
So now, in some straws, double the events means you are above the threshold and you start to mask less straws. The more runs you merge, the less straws you mask.

Here, we load an XML file containing the detectors to mask, as is done in the Zoom workflow.

**Other changes**:

- Add the same mechanism as for ISIS workflows where local files can be used instead of `pooch` files by configuring `DataFolder`
- Only use single files as inputs to `load_nexus` (instead of a file list). If we want to merge runs, we now have to set up a parameter table (added a section in the loki notebook that shows how to do this).

Note that the changes in masks seem to result in slightly worse results at low Q:

Before:
Screenshot at 2024-02-21 16-55-35

After:
Screenshot at 2024-02-21 16-55-42

This may suggest that we need to mask more pixels close to the beam stop? But that would be for the instrument scientist to decide.


src/esssans/types.py
```diff
@@ -153,15 +153,25 @@ class NeXusMonitorName(sciline.Scope[MonitorType, str], str):
     """Name of Incident|Transmission monitor in NeXus file"""
 
 
-class FileList(sciline.Scope[RunType, list], list):
-    """Filenames of BackgroundRun|EmptyBeamRun|SampleRun"""
+PixelMaskFilename = NewType('PixelMaskFilename', str)
```
SimonHeybrock, line 156: Has it been decided that ESS will use ISIS/Mantid pixel mask files? If not, I think this (and related code) should remain in the `esssans.isis` module.

src/esssans/loki/io.py
```diff
@@ -87,46 +106,94 @@ def _merge_runs(data_groups: Iterable[sc.DataGroup], entries: Iterable[str]):
     return out
 
 
+def merge_sample_runs(
```
nvaytet, line 109: I had to specialise the types here for `SampleRun` and `BackgroundRun`, instead of having a single function with `ScatteringRunType`, because if not there would be multiple providers for `LoadedFileContents[RunType]` as the `to_file_contents` below also matches.

I don't think it's pretty. We also need to manually insert the `merge_sample_runs` and `merge_background_runs` into the pipeline if we want to use a parameter table with multiple runs.

Is there a better way to achieve this?
I basically want to merge the runs if a parameter table is supplied, but do nothing otherwise.

src/esssans/types.py
nvaytet: I don't think it's been decided, but I imagine that loading a mask from file would be one of the way the users would want to mask pixels.
I think we do not need to use the Mantid format though, which is why I left the XML details in the `isis` module.
We could use scipp's hdf5 format to store masks, but you'd still have a `PixelMaskFilename` and a `PixelMask`?

src/esssans/loki/io.py
SimonHeybrock: We can ignore this problem for now, as we plan to have a more generic solution.

src/esssans/loki/masking.py
```diff
@@ -3,114 +3,58 @@
...
+    mask.values[np.isin(ids.values, masked.values)] = True
+    return PixelMask(mask)
 
 
-def mask_detectors(
-    da: RawData[ScatteringRunType],
-    lowcounts_straw_mask: Optional[DetectorLowCountsStrawMask],
-    beam_stop_mask: Optional[DetectorBeamStopMask],
-    tube_edge_mask: Optional[DetectorTubeEdgeMask],
+def apply_pixel_masks(
```
SimonHeybrock, line 39: Why is this Loki-specific?

nvaytet: Yes, this is another slightly annoying thing I had to work around.
In `isis`, the masks are applied on `RawDataWithComponentUserOffsets`, while in `loki`, they are applied on `RawData`.

Can we do it differently? I am not sure we want to introduce the `ComponentUserOffsets` everywhere?
I could make the `apply_pixel_masks` a common function and then have wrappers inside `isis` and `loki` that map the correct types, but I feel that would not be ideal either?

SimonHeybrock: What about Skadi etc.? My question was not necessarily if it can be the same for the ISIS instruments, more about all ESS instruments.

nvaytet: Ok, will move into a common place for loki and leave the isis one in the submodule :+1:

SimonHeybrock (approved):

PR #101, opened 2024-02-26
Take into account masks when broadcasting variances

### LINKED ISSUE #89: Take into account pixel masks in upper-bound uncertainty estimate
This will reduce the overestimation.

nvaytet:
Fixes #89

### DIFF
--- a/src/esssans/uncertainty.py
+++ b/src/esssans/uncertainty.py
@@ -48,12 +48,23 @@ def broadcast_with_upper_bound_variances(
     if _no_variance_broadcast(data, template.sizes):
         return data
     data = data.copy()
-    ones = sc.DataArray(data=sc.ones(sizes=template.sizes))
+    broadcast_sizes = {
+        dim: size for dim, size in template.sizes.items() if dim not in data.dims
+    }
+    ones = sc.DataArray(data=sc.ones(sizes=broadcast_sizes))
     if isinstance(template, sc.DataArray):
-        ones.masks.update(template.masks)
-    mult = ones.sum()
-    div = np.prod([data.sizes[dim] for dim in data.dims if dim in template.dims])
-    data.variances *= (mult / div).value
+        broadcast_dims = set(broadcast_sizes)
+        ones.masks.update(
+            {
+                key: mask
+                for key, mask in template.masks.items()
+                if set(mask.dims).issubset(broadcast_dims)
+            }
+        )
+    # mult = ones.sum()
+    # div = np.prod([data.sizes[dim] for dim in data.dims if dim in template.dims])
+    # data.variances *= (mult / div).value
+    data.variances *= ones.sum().value
     return data.broadcast(sizes={**template.sizes, **data.sizes}).copy()
 
 
--- a/src/esssans/uncertainty.py
+++ b/src/esssans/uncertainty.py
@@ -4,6 +4,7 @@
 
 from typing import Dict, TypeVar, Union, overload
 
+import numpy as np
 import scipp as sc
 
 T = TypeVar("T", bound=Union[sc.Variable, sc.DataArray])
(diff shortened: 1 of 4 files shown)


nvaytet:
I'm still unsure what to do about the [broadcast with events](https://github.com/scipp/esssans/blob/main/src/esssans/uncertainty.py#L56)

src/esssans/uncertainty.py
```diff
@@ -4,43 +4,63 @@
...
 def broadcast_with_upper_bound_variances(
-    data: Union[sc.Variable, sc.DataArray], sizes: Dict[str, int]
+    data: Union[sc.Variable, sc.DataArray], template: Union[sc.Variable, sc.DataArray]
 ) -> Union[sc.Variable, sc.DataArray]:
-    if _no_variance_broadcast(data, sizes):
+    """
+    Upper-bound estimate for errors from broadcasting a dense array to match the sizes
+    of a given template array.
+
+    We first count the number of non-masked elements in the template. To get the
```
nvaytet, line 34: I think this approach only works if we are sure that we will subsequently perform a reduction operation along the dimensions that contain masks.

If we never do a reduction, but just compare say a normalization term with `pixel` and `wavelength` dimensions, if some of the pixels were masked, the variances of the normalization term will be different everywhere, compared to a case where no pixels are masked.

This doesn't seem to make much sense conceptually?

SimonHeybrock: Not sure why they would be "different everywhere"?

The upper-bound approach actually ensures that the errors are consistent (and upper bound) for *any* subsequent operation, so even if we never normalize and reduce I think it makes sense?

tests/uncertainty_test.py
```diff
@@ -26,13 +48,145 @@ def test_broadcast_scales_variances_by_new_subspace_volume():
...
+    #                            B B B B B
+    #                            X X X X X
+    #                            X X X X X
+    #                            B B B B B
+    #                            B B B B X
+
+    # TODO: It is not so clear what to do here. This would be, for example, a case where
+    # the direct beam depends on wavelength and layer, while the solid angle depends on
+    # layer and straw, and the mask on the solid angle depends on both layer and straw.
+    # Do we just count the number of non-masked elements in the template?
```
SimonHeybrock, line 138: I think we have to count the combination of masks, i.e., the masks the output would have.

src/esssans/uncertainty.py
```diff
@@ -4,43 +4,63 @@
...
+    if _no_variance_broadcast(data, template.sizes):
         return data
-    size = 1
-    for dim, dim_size in sizes.items():
-        if dim not in data.dims:
-            size *= dim_size
     data = data.copy()
-    data.variances *= size
-    return data.broadcast(sizes={**sizes, **data.sizes}).copy()
+    ones = sc.DataArray(data=sc.ones(sizes=template.sizes))
```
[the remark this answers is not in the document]
nvaytet: > Also we only want to handle pixel masks, not other masks.

Not sure I understand why pixel masks are special? Should we do the same if we have a mask in wavelength or anything else? Unless this is special handling for the SANS workflow, but we never intended to write a generic handling of masks in variance broadcast?

src/esssans/uncertainty.py
SimonHeybrock: > Now say half of the template is masked, we are now only multiplying the variances by 3. If everything stops here, and that is the final result I want to look at, it seems inconsistent?

I think it is consistent. We return a result with no correlations, they have been "absorbed" into the error bars.

src/esssans/uncertainty.py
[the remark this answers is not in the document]
SimonHeybrock: What I mean is, we must only consider masks that relate to the dimensions being broadcasted. Other masks need to be excluded from the counting.

src/esssans/uncertainty.py
```diff
@@ -4,43 +4,63 @@
...
+    if isinstance(template, sc.DataArray):
+        ones.masks.update(template.masks)
```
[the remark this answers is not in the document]
nvaytet: I thought `irreducible_mask` needed to know which `dim` was going to be used for subsequent reduction, but we are potentially broadcasting into multiple new dimensions. So I am not sure what I would use as `dim`?

src/esssans/uncertainty.py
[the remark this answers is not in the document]
nvaytet: > Should make ones only for the broadcast dims, otherwise this is wasteful.

I think this breaks down in the case where the mask is two-dimensional: one of the dims is also in the input, and the other is in the broadcast template?
I guess this may be a unusual case, so I should add special handling.

tests/uncertainty_test.py
nvaytet: So if you had this, you would count only the non-masked pixels in the output?
IMG_0412

nvaytet: But as I wrote in the comment, what do you do if e.g. the direct beam function depends on `[wavelength, layer]`, and the solid angle depends on `[layer, straw, pixel]`, and the mask on the solid angle also depends on `[layer, straw, pixel]`?

SimonHeybrock: Ignore those masks?

nvaytet: Let's talk tomorrow, also about the `irreducible_mask`.

SimonHeybrock:
I have started implementing this in ESSreduce, closing.

PR #104, opened 2024-02-28
Load monitors separately from the remainder of the files

### LINKED ISSUE #99: Load monitors separately from the remainder of the files
This is a subtask of #82, which is easy to implement in isolation. The required monitors should be loaded independently, without loading the rest of the file.

- Load *each* monitor on its own, not all together.
- You can assume that all monitors are direct children of `NXentry/NXinstrument`. The monitor names are defined in the workflow.
- You can assume that position information is self-contained, i.e., there are no `depends_on` chains leading outside a monitor. We can thus directly use `snx.compute_positions` on the individual monitor.
- Add a filter to the existing load function such that no monitors are loaded there. This is essential to avoid potential massive overhead from large event-mode monitors.
- The functions for "extracting" monitors for the loaded data probably need to preserved for supporting legacy NeXus files (ISIS). They should be moved to that submodule if feasible.

jl-wynen:
Fixes #99

I only did this for the Loki NeXus files because we cannot split loading of the ISIS files so easily. And I don't think those matter for performance in the long run.


src/esssans/loki/general.py
```diff
@@ -36,37 +37,31 @@
...
-    out = mon_dg[f'{monitor_name}_events']
-    out.coords['position'] = mon_dg['position']
+    out = monitor[f'{monitor_name}_events'].copy(deep=False)
+    out.coords['position'] = monitor['position']
     return CalibratedMonitor[RunType, MonitorType](out)
 
 
 def detector_pixel_shape(
-    dg: LoadedFileContents[ScatteringRunType], detector_name: NeXusDetectorName
+    dg: LoadedDetector[ScatteringRunType], detector_name: NeXusDetectorName
```
nvaytet, line 55:
```suggestion
    dg: LoadedDetector[ScatteringRunType]
```

src/esssans/loki/io.py
```diff
@@ -79,121 +82,169 @@ def _merge_events(a, b):
 
 def _merge_runs(
     data_groups: sciline.Series[
-        Filename[ScatteringRunType], LoadedSingleFileContents[ScatteringRunType]
+        Filename[ScatteringRunType], LoadedSingleFileDetector[ScatteringRunType]
```
nvaytet, line 85: Below, the name is saying this can be either detector or monitor data, but here we are saying it only applies to detector data?
I know it's just an internal function whose types are not checked when building the pipeline, but I still got a bit confused when reading.

src/esssans/loki/io.py
```diff
@@ -79,121 +82,171 @@ def _merge_events(a, b):
...
     )
-    return LoadedFileContents[BackgroundRun](out)
+    if data_name in DETECTOR_BANK_RESHAPING:
+        events = DETECTOR_BANK_RESHAPING[data_name](events)
+
+    dg[f'{data_name}_events'] = events
+    return dg
 
 
 def load_nexus(
```
nvaytet, line 202:
```suggestion
def load_nexus_detector(
```

src/esssans/loki/io.py
```diff
@@ -79,121 +82,171 @@ def _merge_events(a, b):
...
-    incident_monitor_name: NeXusMonitorName[Incident],
-    transmission_monitor_name: NeXusMonitorName[Transmission],
     transform_path: TransformationPath,
     source_name: NeXusSourceName,
     sample_name: Optional[NeXusSampleName],
-) -> LoadedSingleFileContents[RunType]:
-    data_entries = (detector_name, incident_monitor_name, transmission_monitor_name)
+) -> LoadedSingleFileDetector[RunType]:
+    return LoadedSingleFileDetector[RunType](
+        _load_events(filename, detector_name, transform_path, source_name, sample_name)
```
nvaytet, line 210: Maybe we should use keyword args to make sure we don't mix things up, such as swapping source and sample names?

src/esssans/loki/io.py
```diff
@@ -79,121 +82,171 @@ def _merge_events(a, b):
...
-            data = DETECTOR_BANK_RESHAPING[name](data)
+def load_nexus_monitor(
+    filename: FilePath[Filename[RunType]],
+    monitor_name: NeXusMonitorName[MonitorType],
+    transform_path: TransformationPath,
+    source_name: NeXusSourceName,
+    sample_name: Optional[NeXusSampleName],
+) -> LoadedSingleFileMonitor[RunType, MonitorType]:
+    return LoadedSingleFileMonitor[RunType, MonitorType](
+        _load_events(filename, monitor_name, transform_path, source_name, sample_name)
```
nvaytet, line 222: Also kwargs?

src/esssans/loki/io.py
```diff
@@ -79,121 +82,171 @@ def _merge_events(a, b):
...
 
 def to_path(filename: FilenameType, path: DataFolder) -> FilePath[FilenameType]:
-    return f'{path}/{filename}'
+    return FilePath[FilenameType](f'{path}/{filename}')
 
 
-providers = (load_nexus, to_file_contents, to_path)
+providers = (load_nexus, load_nexus_monitor, to_detector, to_monitor, to_path)
+"""Providers for loading single files."""
+event_merging_providers = (
```
nvaytet, line 246: :+1:

src/esssans/loki/io.py
jl-wynen: A type checker should flag those errors. But I can change to keyword args

nvaytet (approved):

src/esssans/isis/data.py
SimonHeybrock, line 1: @jl-wynen The things moved into this file do not belong here, as it is for the Pooch registry, is in all our repos.

src/esssans/loki/general.py
```diff
@@ -36,37 +37,31 @@
 
 
 def get_detector_data(
-    dg: LoadedFileContents[ScatteringRunType], detector_name: NeXusDetectorName
+    dg: LoadedDetector[ScatteringRunType], detector_name: NeXusDetectorName
 ) -> RawData[ScatteringRunType]:
-    da = dg[NEXUS_INSTRUMENT_PATH][detector_name][f'{detector_name}_events']
-    return RawData[ScatteringRunType](da)
+    return RawData[ScatteringRunType](dg[f'{detector_name}_events'])
```
SimonHeybrock, line 42: Why does this exist?

src/esssans/loki/general.py
```diff
@@ -36,37 +37,31 @@
...
-    dg: LoadedFileContents[RunType], monitor_name: NeXusMonitorName[MonitorType]
+    monitor: LoadedMonitor[RunType, MonitorType],
+    monitor_name: NeXusMonitorName[MonitorType],
 ) -> CalibratedMonitor[RunType, MonitorType]:
-    mon_dg = dg[NEXUS_INSTRUMENT_PATH][monitor_name]
-    out = mon_dg[f'{monitor_name}_events']
-    out.coords['position'] = mon_dg['position']
+    out = monitor[f'{monitor_name}_events'].copy(deep=False)
+    out.coords['position'] = monitor['position']
     return CalibratedMonitor[RunType, MonitorType](out)
```
SimonHeybrock, line 51: Why does this function still exist?

src/esssans/loki/general.py
```diff
@@ -36,37 +37,31 @@
...
 def detector_lab_frame_transform(
-    dg: LoadedFileContents[ScatteringRunType],
-    detector_name: NeXusDetectorName,
+    detector: LoadedDetector[ScatteringRunType],
     transform_path: TransformationPath,
 ) -> LabFrameTransform[ScatteringRunType]:
-    return LabFrameTransform[ScatteringRunType](
-        dg[NEXUS_INSTRUMENT_PATH][detector_name][transform_path]
-    )
+    return LabFrameTransform[ScatteringRunType](detector[transform_path])
```
SimonHeybrock, line 64: Can't all this be done in the detector load function?

SimonHeybrock (commented):
Can't this be simplified further?

PR #108, opened 2024-03-01
Re-organise the doc pages

### LINKED ISSUE #107: Improve documentation
For example, the front page is saying "This documentation is under construction. See the [Sans2d data reduction](https://scipp.github.io/esssans/examples/sans2d.html) example for a quick start.". We need to add a bit more, and maybe think about a first-time user encountering the docs. Where should they start, what should they read? Is there a line they can follow, or are there gaps? Doesn't have to be long, just a couple pointers may be enough.

_Originally posted by @SimonHeybrock in https://github.com/scipp/esssans/issues/6#issuecomment-1970387977_

nvaytet:
- Made a User guide section
- Made folders for Loki, Isis instrument, and Common tools
- Added quick links to those on the index
- Added installation instructions
- Removed the Examples folder

Fixes #107


docs/developer/index.md
```diff
@@ -1,13 +1,11 @@
...
 ```{include} ../../CONTRIBUTING.md
 ```
 
-## Table of contents
-
 ```{toctree}
 ---
-maxdepth: 2
+hidden:
 ---
```
SimonHeybrock, line 9: Why deviate from the common template?

nvaytet: I thought we were allowed to try things out and if we like them we can propagate them to the template?

nvaytet: Fair enough, I put it back in :+1:

SimonHeybrock: I think you edited the wrong file in your latest commit. My comment was about the dev docs, not the general docs.

nvaytet: oops...

SimonHeybrock (approved):

PR #113, opened 2024-03-07
Make namespace package

### LINKED ISSUE #87: Switch to namespace package
To be decided when we can/want to do this.

nvaytet:
Fixes #87


src/ess/isissans/__init__.py
jl-wynen, line 1: Does `isissans` need to be a separate package? Doesn't it make more sense to have `ess.sans.isis`? The isis stuff is temporary anyway, no?

jl-wynen (commented):
Are we ready to drop the support for installing esssans and old ess in the same env?

src/ess/isissans/__init__.py
nvaytet: Yes, I don't really know, I thought because it's temporary then maybe easier to pull it out of `ess.sans`?

As a side note, I named it `isissans` because I thought `from ess import isis` didn't make much sense, since it was just SANS-related stuff.

nvaytet:
> Are we ready to drop the support for installing esssans and old ess in the same env?

I think so? I don't think anyone is using sans and other old ess things at the same time.
I also don't think there is anything left in the old repo we don't have in `esssans`.

src/ess/isissans/__init__.py
jl-wynen: > Yes, I don't really know, I thought because it's temporary then maybe easier to pull it out of ess.sans?

Dunno. I'm fine with it either way.

> As a side note, I named it isissans because I thought from ess import isis didn't make much sense, since it was just SANS-related stuff.

:+1:

jl-wynen (approved):

PR #116, opened 2024-03-20
Try cyclebane

SimonHeybrock:
This demonstrates how Cyclebane could be used to replace param tables, on the example of pixel masks. See the updated Zoom notebook.

To run this, you need:
- Sciline @ `cyclebane` branch
- [Cyclebane](https://github.com/scipp/cyclebane) @ `map-reduce` branch

Note: Tests failing, since I did not refactor those to replace the use of param tables.

### DIFF
--- a/docs/user-guide/isis/zoom.ipynb
+++ b/docs/user-guide/isis/zoom.ipynb
@@ -136,8 +136,7 @@
     "    isis.data.transmission_from_sample_run,\n",
     "    sans.beam_center_from_center_of_mass,\n",
     ")\n",
-    "pipeline = sciline.Pipeline(providers, params=params)\n",
-    "pipeline.set_param_series(PixelMaskFilename, masks)"
+    "pipeline = sciline.Pipeline(providers, params=params)"
    ]
   },
   {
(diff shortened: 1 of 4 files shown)


docs/user-guide/isis/zoom.ipynb
```diff
@@ -175,6 +174,35 @@
...
+    "    graph = pipeline.get(\n",
+    "        IofQ[SampleRun], handler=sciline.HandleAsComputeTimeException()\n",
+    "    )\n",
+    "    cbgraph = cb.Graph(graph.to_networkx())\n",
+    "    cbgraph[MaskedDetectorIDs] = (\n",
+    "        cbgraph[MaskedDetectorIDs]\n",
+    "        .map({PixelMaskFilename: masks})\n",
+    "        .reduce(attrs={\"reduce\": merge})\n",
+    "    )\n",
+    "    return sciline.task_graph.TaskGraph.from_networkx(cbgraph.to_networkx())"
```
jokasimr, line 203: Just to understand what's going on here, is this what it does?
```python
masked_detector_ids = dict()
for pixel_mask_filename in masks:
    pl[PixelMaskFilename] = pixel_mask_filename
    masked_detector_ids = merge(masked_detector_ids, pl.compute(MaskedDetectorIDs))
pl[MaskedDetectorIDs] = masked_detector_ids
```
But I guess it's not the same because in the `sciline-cyclebane` version the execution is delayed.

SimonHeybrock: Yes, I think this is a good way of putting it (if I understand your code correctly, i.e., compute the branch in each case, merge all the result into a single dict, and set it as a value).

PR #119, opened 2024-04-02
Process multiple detector banks via param series

### LINKED ISSUE #83: Load detectors separately and handle multiple detector banks
Loki (and many other instruments) will have multiple detector banks. We need to handle this in a convenient manner. Eventually this will likely also require input from IDS/IS (at least in certain cases), but I believe our current understanding is sufficient to proceed:

- Multiple banks (NXdetector groups in a Nexus file) will be handled using a param table in Sciline.
  - Note: implementation details may change regarding Sciline and param tables, but this is mostly orthogonal to what we need to do here, i.e., we can proceed).
- https://github.com/scipp/esssans/issues/82#issuecomment-1945347421 mentions loading of non-monitor and non-detector data. This will be required in the medium-term future (sample logs, chopper logs, ...), but currently we are not using any of that information. That is, we can essentially ignore this aspect for now, since it can easily be added later.
- Gather results from all banks once we are in $(Q, \lambda)$ or $Q$, but before normalization, `sum` (`concat` if events) them.
  - For now this means inserting an extra step in the workflow, see #97. This may be away of Sciline's way of handling param tables is changed. Do not worry about this for now, just add the step.

Complications, to ignore for now:

- Beam center. See #28. Does it apply only to the rear bank in Loki? Are the others fixed enough, with well-known positions.
- Will we have a different direct-beam for the other banks? This is probably easy to implement, by including direct beam files in the param table, if required

Summary of work to be performed:

- Implement #99 first.
- Instead of loading an entire file, load an individual NXdetector.
  - You can assume that all detectors are direct children of `NXentry/NXinstrument`. The detector names are defined in the workflow, either as a single param `DetectorName` (when processing a single bank), or as a `sciline.ParamTable` (or using https://scipp.github.io/sciline/generated/classes/sciline.Pipeline.html#sciline.Pipeline.set_param_series).
  - You can assume that position information is self-contained, i.e., there are no `depends_on` chains leading outside a detector. We can thus directly use `snx.compute_positions` on the individual detector.
  - Add a filter to the existing load function such that no detectors are loaded there. This is essential to avoid potential massive overhead from large event-mode detectors.
- The functions for "extracting" detectors for the loaded data probably need to preserved for supporting legacy NeXus files (ISIS). They should be moved to that submodule if feasible.
- Add a step for merging intermediate results before performing the normalization. This will require merging the numerator as well as the denominator (the denominator depends on the detector bank via the solid-angle factor). Tracked in https://github.com/scipp/esssans/issues/97.

SimonHeybrock:
Fixes #83.

I could have done the bank and run merge in a single step. Currently it is not clear which option is better in terms of parallelism and memory use, this has to be explored in the future.

~Needs bugfix release of Sciline with scipp/sciline#150.~ (done)


tests/loki/iofq_test.py
```diff
@@ -175,11 +176,49 @@ def test_pipeline_can_compute_IofQ_merging_events_from_multiple_runs():
...
 
 
+def test_pipeline_can_compute_IofQ_merging_events_from_banks():
+    params = make_params()
+    del params[NeXusDetectorName]
+
+    pipeline = sciline.Pipeline(loki_providers_no_beam_center_finder(), params=params)
+    pipeline[BeamCenter] = _compute_beam_center()
+    pipeline.set_param_series(PixelMaskFilename, ['mask_new_July2022.xml'])
+    pipeline.set_param_series(NeXusDetectorName, ['larmor_detector'])
```
nvaytet, line 191: Can we also use e.g. twice the same bank, as I believe for now, only the
```Py
if len(data) == 1:
    return data[0]
```
code branch is being tested, which is a special case?

SimonHeybrock: Unfortunately we cannot, since for a param series the index is the same as the values, and the index must be unique. We could change everything to a param table, introducing an extra type for the row index.

Note that the other branch (of the helper function) is tested by the run-merge. But I agree that this is not ideal. I will try to write a unit test just for the merge function.

nvaytet (approved):

SimonHeybrock:
Will wait with merging until we can update to bugfix release of Sciline (for graph grouping).

PR #123, opened 2024-04-08
Example workflow for Loki to enable beamlime

MridulS:
This setups a workflow to address https://github.com/scipp/beamlime/issues/159.

Not too sure where to keep the function but for now I have created a new file.

Fixes scipp/beamlime#159.

### DIFF
--- a/src/ess/loki/__init__.py
+++ b/src/ess/loki/__init__.py
@@ -5,7 +5,6 @@ import importlib.metadata
 
 from . import data, general, io
 from .general import default_parameters
-from .workflow import live_workflow
 
 try:
     __version__ = importlib.metadata.version(__package__ or __name__)
@@ -16,4 +15,4 @@ providers = general.providers + io.providers
 
 del importlib
 
-__all__ = ['data', 'general', 'io', 'providers', 'default_parameters', 'live_workflow']
+__all__ = ['data', 'general', 'io', 'providers', 'default_parameters']
--- a/src/ess/loki/__init__.py
+++ b/src/ess/loki/__init__.py
@@ -5,6 +5,7 @@ import importlib.metadata
 
 from . import data, general, io
 from .general import default_parameters
+from .workflow import live_workflow
 
 try:
     __version__ = importlib.metadata.version(__package__ or __name__)
@@ -15,4 +16,4 @@ providers = general.providers + io.providers
 
 del importlib
 
-__all__ = ['data', 'general', 'io', 'providers', 'default_parameters']
+__all__ = ['data', 'general', 'io', 'providers', 'default_parameters', 'live_workflow']
--- a/src/ess/loki/workflow.py
+++ b/src/ess/loki/workflow.py
@@ -1,6 +1,5 @@
 # This file is used by beamlime to create a workflow for the Loki instrument.
 # The function `live_workflow` is registered as the entry point for the workflow.
-import numpy as np
 import scipp as sc
 from beamlime import WorkflowResult
 from scippneutron.io.nexus.load_nexus import JSONGroup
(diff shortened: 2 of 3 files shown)


MridulS:
What's the policy about adding dependencies like beamlime here? Should I just add it as a proper dependency or as an optional one as it is only required by a specific code path.

SimonHeybrock:
> What's the policy about adding dependencies like beamlime here? Should I just add it as a proper dependency or as an optional one as it is only required by a specific code path.

We have not actually discussed this specific case. The live-data dashboard will not be run by regular users, but "by us" on a dedicated machine (VM), so personally I would think we can simply *not* add beamlime as a dependency (and install by hand when needed).

src/ess/loki/__init__.py
```diff
@@ -5,6 +5,7 @@
 
 from . import data, general, io
 from .general import default_parameters
+from .workflow import live_workflow
```
SimonHeybrock, line 8: Can we avoid the import, such that users can still import this module without having to install Beamlime (see my earlier comment)?

src/ess/loki/workflow.py
```diff
@@ -0,0 +1,24 @@
+try:
+    import numpy as np
+    import scipp as sc
+    from beamlime import WorkflowResult
+    from scippneutron.io.nexus.load_nexus import JSONGroup
+except ImportError as e:
+    raise ImportError(
+        "Please install scipp, scippneutron, and beamlime to run workflows with loki."
```
SimonHeybrock, line 8: The former two are actually in the dependencies already.

src/ess/loki/workflow.py
```diff
@@ -0,0 +1,24 @@
...
+    """Example live workflow function for Loki.
+
+    This function is used to process data with beamlime workflows.
+    """
+    # We do not consume the incoming data right now.
+    _ = group
+    return {
+        'IofQ': sc.DataArray(
+            data=sc.array(dims=['Q'], values=np.zeros(10)),
+            coords={'Q': sc.array(dims=['Q'], values=np.zeros(10), unit='1/angstrom')},
```
SimonHeybrock, line 22: I'd make this a `sc.linspace` with reasonable bounds, so plotting will work (for example as used in the some of the examples in the docs: `sc.linspace(dim='Q', start=0.01, stop=0.3, num=101, unit='1/angstrom')`).

MridulS:
> (and install by hand when needed).

Maybe keep this is an optional dependency then? Could be easier to track all deps required for live-data dashboard (if and when more are added) in a single place. Something like `pip install esssans[live]`

src/ess/loki/workflow.py
```diff
@@ -0,0 +1,25 @@
...
+def live_workflow(group: JSONGroup) -> WorkflowResult:
+    """Example live workflow function for Loki.
+
+    This function is used to process data with beamlime workflows.
+    """
+    # We do not consume the incoming data right now.
+    _ = group
+    return {
+        'IofQ': sc.DataArray(
+            data=sc.array(dims=['Q'], values=np.zeros(101)),
```
SimonHeybrock, line 18: This should actually be length 100, since Q is typically bin-edges in our applications.

SimonHeybrock: Note there is also `sc.zeros`, which avoids Numpy in this case.

src/ess/loki/workflow.py
```diff
@@ -15,10 +14,10 @@ def live_workflow(group: JSONGroup) -> WorkflowResult:
     _ = group
     return {
         'IofQ': sc.DataArray(
-            data=sc.array(dims=['Q'], values=np.zeros(101)),
+            data=sc.zeros(dims=['Q'], shape=[100]),
             coords={
                 'Q': sc.linspace(
-                    dim='Q', start=0.01, stop=0.3, num=101, unit='1/angstrom'
+                    dim='Q', start=0.01, stop=0.3, num=100, unit='1/angstrom'
```
SimonHeybrock, line 20: Sorry, I was unclear: The data has length 100, but the coord length 101, i.e., this is a histogram where `Q` gives the bin edges.

SimonHeybrock (approved):

SimonHeybrock:
> > (and install by hand when needed).
> 
> Maybe keep this is an optional dependency then? Could be easier to track all deps required for live-data dashboard (if and when more are added) in a single place. Something like `pip install esssans[live]`

You are right, this might be what we do in the long run. But as Beamlime (and in particular this plugin mechanism) is in a very early state we might wait a little.

PR #125, opened 2024-04-12
fix: add associated monitor workspace to data workspace

jokasimr:
This change makes the ISISSans workflow able to handle files with periods.
Annika has such files where the data is separated into periods based on polarization settings.

### DIFF
--- a/src/ess/isissans/mantidio.py
+++ b/src/ess/isissans/mantidio.py
@@ -10,7 +10,14 @@ import scipp as sc
 import scippneutron as scn
 from scipp.constants import g
 
-from ..sans.types import DirectBeam, DirectBeamFilename, Filename, RunType, SampleRun, Period
+from ..sans.types import (
+    DirectBeam,
+    DirectBeamFilename,
+    Filename,
+    Period,
+    RunType,
+    SampleRun,
+)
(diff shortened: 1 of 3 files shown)


src/ess/sans/types.py
```diff
@@ -352,3 +352,6 @@ class CleanMonitor(
 
 RunNumber = NewType('RunNumber', int)
 """Run number."""
+
+Period = NewType('Period', int)
+"""Period number."""
```
SimonHeybrock, line 357: Periods are an ISIS thing, this should only be defined in that submodule.

src/ess/isissans/mantidio.py
```diff
@@ -106,6 +115,13 @@ def load_run(filename: FilePath[Filename[RunType]]) -> DataWorkspace[RunType]:
     else:
         # Separate data and monitor workspaces
         data_ws = loaded.OutputWorkspace
+        if isinstance(data_ws, _mantid_api.WorkspaceGroup):
+            if period is None:
+                raise ValueError('Needs period to know what section to load')
```
SimonHeybrock, line 120: Could be more specific about which parameter needs to be set.

SimonHeybrock (approved):

Issue #95, opened 2024-02-20
Use meaningful domain type locations/imports


SimonHeybrock:
Currently many "types" are in `esssans.types`, and we use `from esssans.types import *`. Apart form wanting to avoid the `import *`, `types` is also meaningless to scientists. Therefore:

- Move important types to the package root (import in `__init__.py`), avoid `import *`.
- More special or internal types should move to meaningful submodules.

SimonHeybrock:
Why would you use `Filename` from `iotypes`? `Filename` should be imported in the top-level `__init__.py`, since it is commonly used by users.

SimonHeybrock:
To reiterate form previous discussions:

- Define where it makes sense (use your judgement), might be in local file defining providers.
- `types` should not be exposed to users.
- "Relevant" types get imported in top-level `__init__.py`.
- Details have to be imported from their submodules.

nvaytet:
> Filename should be imported in the top-level __init__.py, since it is commonly used by users.

I've tried a couple of things, but I can't seem to get around having to have a long
```Py
from esssans import (
    Filename,
    SampleRun,
    BackgroundRun,
    TransmissionRun,
    EmptyBeamRun,
    WavelengthBins,
    WavelengthBands,
    CorrectForGravity,
    UncertaintyBroadcastMode,
    ReturnEvents,
    QBins,
    PixelMaskFilename,
    BackgroundSubtractedIofQ,
)
```
at the start of the notebook. Was that the intention?

Another thing: without the `import *`, if I want to inspect an intermediate result that I look up in the task graph, I now have to import that type before I can look at the data
```Py
from esssans.types import MaskedData
masked = pipeline.compute(MaskedData[SampleRun])
```
which degrades usability I think.

I know we don't like the `import *`, but I feel the aternative should not lower usability this much?

SimonHeybrock:
It is your choice to `import *`, but I don't feel we should adopt that practice in the docs?

nvaytet:
> It is your choice to `import *`

Sure, but that becomes very difficult if the types are scattered over many files/submodules internally.

So unless we find a better alternative, I might suggest to have, for every submodule, a single place where we put all the types (and we don't have to call it types if you think the name doesn't make sense to scientists).

This would allow you to do:
```Py
from esssans.types import *
from esssans.loki.types import *
```
and you would have all you need if you are working with Loki. We would then not have any types in the other files in the `loki` submodule.

[...]

PR #127, opened 2024-04-16
MAINT: Test using ruff with minimal changes to defaults

MridulS:
This adds an example ruff config which uses all the defaults with minimal config. It makes very little changes.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,15 +13,6 @@ repos:
       - id: trailing-whitespace
         args: [ --markdown-linebreak-ext=md ]
         exclude: '\.svg'
-  - repo: https://github.com/pycqa/isort
-    rev: 5.12.0
-    hooks:
-      - id: isort
-        name: isort (python)
-  - repo: https://github.com/psf/black-pre-commit-mirror
-    rev: 23.11.0
-    hooks:
-      - id: black
   - repo: https://github.com/kynan/nbstripout
     rev: 0.6.0
     hooks:
@@ -29,18 +20,6 @@ repos:
         types: [ "jupyter" ]
         args: [ "--drop-empty-cells",
                 "--extra-keys 'metadata.language_info.version cell.metadata.jp-MarkdownHeadingCollapsed cell.metadata.pycharm'" ]
-  - repo: https://github.com/pycqa/flake8
-    rev: 6.1.0
-    hooks:
-      - id: flake8
-        types: ["python"]
-        additional_dependencies: ["flake8-bugbear==23.9.16"]
-  - repo: https://github.com/pycqa/bandit
-    rev: 1.7.5
-    hooks:
-      - id: bandit
-        additional_dependencies: ["bandit[toml]"]
-        args: ["-c", "pyproject.toml"]
   - repo: https://github.com/codespell-project/codespell
     rev: v2.2.6
     hooks:
@@ -57,3 +36,11 @@ repos:
       - id: rst-directive-colons
       - id: rst-inline-touching-normal
       - id: text-unicode-replacement-char
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.3.7
+    hooks:
+      - id: ruff
+        types_or: [ python, pyi, jupyter ]
+        args: [ --fix]
+      - id: ruff-format
+        types_or: [ python, pyi]
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,5 +1,3 @@
-# -*- coding: utf-8 -*-
-
 import doctest
 import os
 import sys
@@ -8,9 +6,9 @@ from importlib.metadata import version as get_version
 sys.path.insert(0, os.path.abspath('.'))
 
 # General information about the project.
-project = u'ESSsans'
-copyright = u'2024 Scipp contributors'
-author = u'Scipp contributors'
+project = 'ESSsans'
+copyright = '2024 Scipp contributors'
+author = 'Scipp contributors'
 
 html_show_sourcelink = True
 
--- a/docs/user-guide/isis/sans2d.ipynb
+++ b/docs/user-guide/isis/sans2d.ipynb
@@ -22,11 +22,11 @@
    },
    "outputs": [],
    "source": [
-    "import scipp as sc\n",
-    "import sciline\n",
     "import plopp as pp\n",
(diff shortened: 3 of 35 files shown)


MridulS:
> does not use rules based on isort, bugbear, and bandit. So we still need to run those linters.

urghh, for some reason I thought the default was running all the rules (it's not). Need to select ALL.

docs/conf.py
```diff
@@ -1,39 +1,37 @@
-# -*- coding: utf-8 -*-
-
 import doctest
 import os
 import sys
 from importlib.metadata import version as get_version
 
-sys.path.insert(0, os.path.abspath('.'))
+sys.path.insert(0, os.path.abspath("."))
```
jl-wynen, line 6: We don't enforce quotation styles. We had discussions about this and could not agree. So we just do it manually.

pyproject.toml
```diff
@@ -65,17 +65,19 @@ filterwarnings = [
   'ignore:\n            Sentinel is not a public part of the traitlets API:DeprecationWarning',
 ]
 
-[tool.bandit]
-# Excluding tests because bandit doesn't like `assert`.
-exclude_dirs = ["docs/conf.py", "tests"]
+[tool.ruff.lint]
+select = ["I", "UP", "B", "S"]
```
jl-wynen, line 69: Still less than in scipp. In particular, I would like to use `T20`, `PGH`, `RUF` in order to enforce proper use of `noqa` and `type: ignore` directives. (Among some other nice checks.)

But more importantly, this deselects `F` which is the default flake8-based lints.

MridulS: Yup, I am playing around with some more rules to try to get a generalised one.

Also for some reason ruff think scipp is pandas and tries to run https://docs.astral.sh/ruff/rules/#pandas-vet-pd on the codebase (enforcing `.values` to `.to_numpy`)

jl-wynen: > Also for some reason ruff think scipp is pandas and tries to run https://docs.astral.sh/ruff/rules/#pandas-vet-pd on the codebase (enforcing .values to .to_numpy)

This may be an example of a checker that cannot safely be used in all codebases. As I understand it, ruff simply inspects the local parts of the AST but does not do any type checking. So it cannot know whether it is dealing with a pandas data frame or something else.

.pre-commit-config.yaml
```diff
@@ -13,47 +13,24 @@ repos:
...
           - tomli
-  - repo: https://github.com/pre-commit/pygrep-hooks
-    rev: v1.10.0
-    hooks:
-      - id: python-no-eval
-      - id: python-no-log-warn
-      - id: python-use-type-annotations
-      - id: rst-backticks
-      - id: rst-directive-colons
-      - id: rst-inline-touching-normal
```
MridulS, line 58: I don't think we are even using rst for documenting things?
The other rules should be covered by other ruff rules.

pyproject.toml
```diff
@@ -31,25 +28,47 @@ requires-python = ">=3.10"
...
 [tool.setuptools_scm]
 
+[tool.ruff.lint]
+select = ["B", "C4", "DTZ", "E",  "F", "FA",
+          "G", "I", "S", "T20", "PERF", "PGH",
+          "PT", "PYI", "RUF", "UP", "W"]
+# FIXME: This shouldn't be required but tests
+# fail when PEP 585 is followed
+ignore = ['UP006', 'UP035', 'UP007']
+pydocstyle.convention = "numpy"
```
MridulS, line 58: This needs to be fixed.

pyproject.toml
```diff
@@ -31,25 +28,47 @@ requires-python = ">=3.10"
...
+
+[tool.ruff.lint.per-file-ignores]
+"__init__.py" = ["I"]
+# Assert in tests are fine
+"tests/*" = ["S101"]
+# star imports to import types in example notebooks.
+# Do we want to use star imports?
+"*.ipynb" = ["F403", "F405"]
```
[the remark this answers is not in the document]
jl-wynen: We had discussions about this. E.g. here https://github.com/scipp/esssans/issues/95 There has not been a hard decision. But for usability, the star import is hard to beat.

pyproject.toml
jl-wynen: What are you referring to? The upgrade lints? They could be done right now.

The docstyle setting should do nothing right now because `D`-rules are not enabled.

.pre-commit-config.yaml
jl-wynen: We are inside of docstrings. Just not in separate files. And we have had many cases where those rules would have been useful before we enabled them.

Plus `text-unicode-replacement-char` runs on all text files while ruff only runs on python files.

.pre-commit-config.yaml
```diff
@@ -13,47 +13,24 @@ repos:
...
-      - id: text-unicode-replacement-char
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.3.7
+    hooks:
+      - id: ruff-format
+        types_or: [ python, pyi]
+      - id: ruff
+        types_or: [ python, pyi, jupyter ]
+        args: [ --fix]
```
jl-wynen, line 36: The docs recommend first running the linter, then the formatter. I changed it in scipp. Can you do this here, too?

pyproject.toml
MridulS: > What are you referring to? The upgrade lints? They could be done right now.

It fails the test for some reason! I will dig into this a bit later.

docs/user-guide/isis/sans2d.ipynb
```diff
@@ -248,7 +247,7 @@
     ")\n",
     "parts = (CleanSummedQ[SampleRun, Numerator], CleanSummedQ[SampleRun, Denominator])\n",
     "iofqs = (IofQ[SampleRun], IofQ[BackgroundRun], BackgroundSubtractedIofQ)\n",
-    "keys = monitors + (MaskedData[SampleRun],) + parts + iofqs\n",
+    "keys = (*monitors, MaskedData[SampleRun], *parts, *iofqs)\n",
```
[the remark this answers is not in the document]
jl-wynen: It's more efficient. Plus it avoids the awkward `(x,)` syntax.

MridulS:
I'll close this one, once https://github.com/scipp/scipp/pull/3424 is merged in we should create one in the template and let the github action do the work.

PR #132, opened 2024-04-23
Avoid `Optional` in cases where it is not necessary

SimonHeybrock:
This is a subset of the more controversial changes in scipp/esssans#130. I *think* these are probably good cases where `Optional` can be avoided in any case?

Fixes scipp/ess#368.

### DIFF
--- a/tests/loki/directbeam_test.py
+++ b/tests/loki/directbeam_test.py
@@ -39,7 +39,7 @@ def test_can_compute_direct_beam_for_all_pixels():
     providers = loki_providers()
     pipeline = sciline.Pipeline(providers, params=params)
     pipeline.set_param_series(PixelMaskFilename, [])
-    I0 = _get_I0(qbins=params[QBins].edges['Q'])
+    I0 = _get_I0(qbins=params[QBins])
 
     results = sans.direct_beam(pipeline=pipeline, I0=I0, niter=4)
     iofq_full = results[-1]['iofq_full']
@@ -69,7 +69,7 @@ def test_can_compute_direct_beam_with_overlapping_wavelength_bands():
     providers = loki_providers()
     pipeline = sciline.Pipeline(providers, params=params)
     pipeline.set_param_series(PixelMaskFilename, [])
-    I0 = _get_I0(qbins=params[QBins].edges['Q'])
+    I0 = _get_I0(qbins=params[QBins])
 
     results = sans.direct_beam(pipeline=pipeline, I0=I0, niter=4)
     iofq_full = results[-1]['iofq_full']
@@ -95,7 +95,7 @@ def test_can_compute_direct_beam_per_layer():
     providers = loki_providers()
     pipeline = sciline.Pipeline(providers, params=params)
     pipeline.set_param_series(PixelMaskFilename, [])
-    I0 = _get_I0(qbins=params[QBins].edges['Q'])
+    I0 = _get_I0(qbins=params[QBins])
 
     results = sans.direct_beam(pipeline=pipeline, I0=I0, niter=4)
     iofq_full = results[-1]['iofq_full']
@@ -123,7 +123,7 @@ def test_can_compute_direct_beam_per_layer_and_straw():
     providers = loki_providers()
     pipeline = sciline.Pipeline(providers, params=params)
     pipeline.set_param_series(PixelMaskFilename, [])
-    I0 = _get_I0(qbins=params[QBins].edges['Q'])
+    I0 = _get_I0(qbins=params[QBins])
 
     results = sans.direct_beam(pipeline=pipeline, I0=I0, niter=4)
     iofq_full = results[-1]['iofq_full']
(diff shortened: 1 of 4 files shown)


src/ess/sans/types.py
```diff
@@ -119,11 +119,31 @@ class TransmissionRun(sciline.Scope[ScatteringRunType, int], int):
...
 
 
-QBins = NewType('QBins', sc.Variable)
-"""Q binning"""
+class QBins:
+    """Q binning, either 1-D (Q) or 2-D (Qx,Qy)."""
+
+    def __init__(
+        self,
+        bins: sc.Variable | None = None,
```
nvaytet, line 127:
```suggestion
        Q: sc.Variable | None = None,
```
? to be consistent with the others?

tests/loki/directbeam_test.py
```diff
@@ -32,7 +38,8 @@ def test_can_compute_direct_beam_for_all_pixels():
     )
     providers = loki_providers()
     pipeline = sciline.Pipeline(providers, params=params)
-    I0 = _get_I0(qbins=params[QBins])
+    pipeline.set_param_series(PixelMaskFilename, [])
+    I0 = _get_I0(qbins=params[QBins].edges['Q'])
```
nvaytet, line 42: I can see that this is not optimal UX. But at the same time, I don't really know how else to do this. The `QBins` is serving a double purpose. It reminds me a little of the `WavelengthBins` vs `WavelengthBands` which are kind of similar but not really the same thing.

Are `QxyBins` and `QBins` actually mutually exclusive? I know that is what we said in the past. But could you ever want to make a plot of e.g. `Qx` vs `Q`?

Could another alternative be to have a parameter for each `QBins`, `QxBins`, `QyBins` and then you have to request a different result, e.g. `BackgroundSubtractedIofQ` vs `BackgroundSubtractedIofQxy`. And then you would have different providers that take in either `Qbins` or `QxBins` and `QyBins`?

src/ess/sans/types.py
SimonHeybrock: It is a position-only arg, so the name does not matter (cannot be used as a keyword arg).

tests/loki/directbeam_test.py
SimonHeybrock: > Are `QxyBins` and `QBins` actually mutually exclusive? I know that is what we said in the past. But could you ever want to make a plot of e.g. `Qx` vs `Q`?

I don't think so, and catering to hypotheticals does not feel like a good choice?
 
> Could another alternative be to have a parameter for each `QBins`, `QxBins`, `QyBins` and then you have to request a different result, e.g. `BackgroundSubtractedIofQ` vs `BackgroundSubtractedIofQxy`. And then you would have different providers that take in either `Qbins` or `QxBins` and `QyBins`?

Hmmm, interesting idea, I wonder how many providers and types would need to be duplicated, or whether generics could be used.

nvaytet: I like the generics because it looks like a mathematical function `I[Q]`.
Would we then have `CleanSummed[Qxy, SampleRun, Numerator]`?
Could we have a problem where we have too many params in a generic and things start becoming difficult? I can't think a particular example right now but I seem to remember having issues with that (maybe it was when also mixing with parameter tables?).
In any case, if it works, I think it could be nice?

SimonHeybrock: Are we happy to make the change to such generics now, alongside this change? It would represent a major workflow change (for the users).

Also, it will not work with the current release of Sciline, since we do not have a `ScopeThreeParams` helper.

src/ess/sans/normalization.py
```diff
@@ -413,14 +414,55 @@ def _reduce(da: sc.DataArray) -> sc.DataArray:
     numerator /= denominator.drop_coords(
         [name for name in denominator.coords if name not in denominator.dims]
     )
-    return IofQ[ScatteringRunType](numerator)
+    return numerator
+
+
+def normalize_q(
```
nvaytet, line 420: I was wondering for all of these if it would be useful to use `functools.wraps` to copy over the docstring of the internal `_normalize` implementation?
But then it would also update the signature, which would replace the input types `FinalSummedQ[ScatteringRunType, Numerator]` with `sc.DataArray`, which is maybe not something we want? (unless there is a way to copy only the docstring over?)

tests/loki/directbeam_test.py
```diff
@@ -32,6 +38,7 @@ def test_can_compute_direct_beam_for_all_pixels():
     )
     providers = loki_providers()
     pipeline = sciline.Pipeline(providers, params=params)
+    pipeline.set_param_series(PixelMaskFilename, [])
```
nvaytet, line 41: Is it possible to have a default param series so one does not have to specify this every time? (I mean like the `default_parameters`). I guess not...

src/ess/sans/normalization.py
SimonHeybrock: Yes, I had the same thought and decided against it.

tests/loki/directbeam_test.py
SimonHeybrock: Not possible currently, but this is changing anyway with the Sciline rewrite.

nvaytet (approved):

PR #134, opened 2024-04-26
Monitor wavelength binning workflow example.

YooSunYoung:
Fixes scipp/beamlime#186

### DIFF
--- a/src/ess/loki/workflow.py
+++ b/src/ess/loki/workflow.py
@@ -1,7 +1,8 @@
 # This file is used by beamlime to create a workflow for the Loki instrument.
-# The function `live_workflow` is registered as the entry point for the workflow.
+# The callable `live_workflow` is registered as the entry point for the workflow.
 import sciline
 import scipp as sc
+import scippnexus as snx
 from scippneutron.io.nexus.load_nexus import JSONGroup
 
 from ess.loki.general import (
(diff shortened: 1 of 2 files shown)


YooSunYoung:
@SimonHeybrock Will you check if it's in the right direction...?

YooSunYoung:
This is what I tried to make:

workflow

src/ess/loki/workflow.py
```diff
@@ -1,24 +1,89 @@
...
-    return {
-        'IofQ': sc.DataArray(
-            data=sc.zeros(dims=['Q'], shape=[100]),
-            coords={
-                'Q': sc.linspace(
-                    dim='Q', start=0.01, stop=0.3, num=101, unit='1/angstrom'
-                )
-            },
-        )
-    }
```
SimonHeybrock, line 24: We should keep the IofQ workflow, since we will still need this for testing. Add a new one?

src/ess/loki/workflow.py
```diff
@@ -1,24 +1,89 @@
...
+class BinnedMonitor(
+    sciline.ScopeTwoParams[RunType, MonitorType, sc.DataArray], sc.DataArray
+):
+    ...
+
+
+def _bin_monitor_wavelength(
+    wavelength_bin: WavelengthBins, monitor: WavelengthMonitor[RunType, MonitorType]
+) -> BinnedMonitor[RunType, MonitorType]:
+    return monitor.bin(wavelength=wavelength_bin)
```
SimonHeybrock, line 37: Should we actually `hist`? Visualization should not have to do this itself, I think?

src/ess/loki/workflow.py
```diff
@@ -1,24 +1,89 @@
...
+        LoadedNeXusMonitor[SampleRun, Incident]: load_nexus_monitor(group, "monitor_1"),
+        LoadedNeXusMonitor[SampleRun, Transmission]: load_nexus_monitor(
+            group, "monitor_2"
+        ),
+        # Above lines should be replaced with the following commented code
+        # when the `sciline` library is updated to allow setting
+        # parameters with instances of different types than the keys.
+        # FilePath[Filename[SampleRun]]: group,
+        # NeXusMonitorName[Incident]: "monitor_1",
+        # NeXusMonitorName[Transmission]: "monitor_2",
```
SimonHeybrock, line 70: ... and once that is done we can consider moving thie pipeline init to a constructor?

YooSunYoung:
I put the IofQ back.

stateless_workflow

src/ess/loki/workflow.py
```diff
@@ -1,24 +1,139 @@
...
+                    dim="Q", start=0.01, stop=0.3, num=101, unit="1/angstrom"
                 )
             },
         )
-    }
+    )
+
+
+def _hist_iofq(i_of_q: IofQ[ScatteringRunType]) -> IofQHistogram[ScatteringRunType]:
+    return IofQHistogram[ScatteringRunType](i_of_q.hist("Q"))
```
SimonHeybrock, line 62: `hist` does not accept string argument, I think. Why is this code part of this PR?

src/ess/loki/workflow.py
```diff
@@ -1,24 +1,139 @@
...
+            group, definitions=snx.base_definitions()
+        )
+        results = self.pipeline.compute(
+            (
+                MonitorHistogram[SampleRun, Incident],
+                MonitorHistogram[SampleRun, Transmission],
+            )
+        )
+
+        return {str(tp): result for tp, result in results.items()}
```
SimonHeybrock, line 136: Missing `IofQ`? You could keep that workflow separate, and add this one as a second plugin.

YooSunYoung:
Since the workflow accepts `JsonGroup`, how should they be tested...?
Is there any tools to convert `NexusGroup` into `JsonGroup`...?

SimonHeybrock (commented):
Has this been tested with the latest `Beamlime`? Do we get data through the entire chain?

YooSunYoung:
> Has this been tested with the latest `Beamlime`? Do we get data through the entire chain?

I tried but I couldn't make it work with the `loki.json` template.
It's complaining with this error message: 
```
ValueError: Raw data loaded from NeXus does not contain events or a histogram. Expected to find a data array, but the data only contains {'depends_on', 'name', 'transformations', 'transform', 'monitor_2_events', 'position'}
```
That's why I asked this:
> Since the workflow accepts JsonGroup, how should they be tested...?
> Is there any tools to convert NexusGroup into JsonGroup...?

YooSunYoung:
> Has this been tested with the latest `Beamlime`? Do we get data through the entire chain?

The minitor workflow works with the fixes that I mentioned here: https://github.com/scipp/beamlime/issues/193 

The fixes are stashed in this branch: https://github.com/scipp/beamlime/tree/atleast-work

I'll fix them in an order...

I got these plots as a result :D
test-ess loki workflow MonitorHistogram ess sans types SampleRun, ess sans types Incident
test-ess loki workflow MonitorHistogram ess sans types SampleRun, ess sans types Transmission

SimonHeybrock (approved):

PR #135, opened 2024-04-30
Refactor for next Sciline (Optional, param tables, and more)

### LINKED ISSUE #97: Explicit merge / accumulation strategies and customization points
### Overview

Over the past months it has become very cleat that workflow need to be able to merge data in various ways:

- Multiple files that conceptually belong to a single run, first implemented in #35.
- Chunks (wall-clock time-slices) from a single very large file.
- Detector banks (see #83).
- Chunks when streaming data (see initial interface discussions in scipp/beamlime#131)

It is important to realize that these are recurrent requirements, spanning file-based and stream-based workflows, and this is a generic problem that will affect all techniques/instruments.

At the same time, the merge / accumulate operations are not always performed, and it is not clear in advance at which point in a workflow a scientist will want/need to merge. We also have to ensure that merging is easy or almost transparent for scientists, without requiring in-depth knowledge, or the ability to modify workflows in a complicated manner.

I therefore propose to add customization points at which merges can be performed into workflows. I thing typically there may be a handful of such points, e.g., (1) concat right after loading, (2) concat/sum before normalization, (3) average after normalization.

- By default no-op accumulation providers (taking only a single input, not a `sciline.Series`) would be part of the workflow.
- The package would provide a set of commonly used accumulators suitable for insertion at these points. These would, e.g., take `sciline.Series` as input, and require replacing a param by a param table.
- Multiple accumulation points may be used simultaneously, e.g., to merge banks before normalization and average over runs after normalization. In this case there would be two parameter tables.

### Open points

My hope is that by making this explicit in the workflow, we can use that functionality for the streaming case, without having to write complicated or custom just for the sake of live-reduction. The details require some more thought.

### References

Related: #49, scipp/esspolarization#19, scipp/beamlime#131

### Example

This is not complete, but I hope it helps illustrating the point:

```python
import sciline
from typing import NewType

Bank = NewType('Bank', str)
BankId = NewType('BankId', str)
Run = NewType('Run', int)
RunId = NewType('RunId', int)
RawData = NewType('RawData', int)
RawMonitor = NewType('RawMonitor', int)
Counts = NewType('Counts', int)
AccumulatedCounts = NewType('AccumulatedCounts', int)
Norm = NewType('Norm', float)
Normalized = NewType('Normalized', float)
AverageNormalized = NewType('AverageNormalized', float)

def load(run: RunId, bank: BankId) -> RawData:
    return RawData(42)

def load_monitor(run: RunId) -> RawMonitor:
    return RawMonitor(42)

def counts(data: RawData) -> Counts:
    return Counts(data)

def no_accum_counts(counts: Counts) -> AccumulatedCounts:
    return AccumulatedCounts(counts)

def accum_counts(counts: sciline.Series[Run, Counts]) -> AccumulatedCounts:
    return AccumulatedCounts(sum(counts.values()))

def norm_term(mon: RawMonitor) -> Norm:
    return Norm(mon)

def normalize(counts: AccumulatedCounts, norm: Norm) -> Normalized:
    return Normalized(counts / norm)

def no_average_normalized(normalized: Normalized) -> AverageNormalized:
    return AverageNormalized(normalized)

def average_normalized(normalized: sciline.Series[Run, Normalized]) -> AverageNormalized:
    return AverageNormalized(sum(normalized.values()) / len(normalized))

providers = (load, load_monitor, counts, norm_term, normalize)
pipeline = sciline.Pipeline(providers)
runs = sciline.ParamTable(Run, {RunId: range(3)})
banks = sciline.ParamTable(Bank, {BankId: ['A', 'B']})
pl = pipeline.copy()

pipeline[RunId] = 12345
pipeline[BankId] = 'A'
pipeline.insert(no_accum_counts)
pipeline.insert(no_average_normalized)
display(pipeline.visualize(AverageNormalized, graph_attr={'rankdir': 'LR'}))

pipeline.set_param_table(runs)
pipeline.insert(accum_counts)
display(pipeline.visualize(AverageNormalized, graph_attr={'rankdir': 'LR'}))

pipeline.insert(no_accum_counts)
pipeline.insert(average_normalized)
display(pipeline.visualize(AverageNormalized, graph_attr={'rankdir': 'LR'}))

def accum_counts_from_banks(banks: sciline.Series[Bank, Counts]) -> AccumulatedCounts:
    return AccumulatedCounts(sum(banks.values()))

pipeline.set_param_table(banks)
pipeline.insert(accum_counts_from_banks)
pipeline.insert(average_normalized)
display(pipeline.visualize(AverageNormalized, graph_attr={'rankdir': 'LR'}, compact=False))
```
image

### LINKED ISSUE #142: Refactor for next Sciline release

SimonHeybrock:
Changes:
- Add helpers such as `set_banks`, replacing `Pipeline.set_param_table`.
- Remove complicated multi-step helpers for run and bank merges.
- Small changes to work with new way of `Optional` handling by Sciline.
- Change `masking.py` to work more cleanly without param table with a small local workflow structure change.

Notes:
- Depends on Sciline@main, need to update deps once released.
- ~Depends on fixes in Cyclebane which are not even in it's `main` yet.~

Fixes #97.
Fixes #142.


docs/user-guide/common/beam-center-finder.ipynb
```diff
@@ -58,7 +58,10 @@
     "providers = list(\n",
     "    sans.providers + isis.providers + isis.data.providers + isis.sans2d.providers\n",
     ")\n",
-    "pipeline = sciline.Pipeline(providers, params=params)"
+    "pipeline = sciline.Pipeline(providers, params=params)\n",
```
nvaytet, line 61: I guess some of the changes here clash with #136.
What is the merging order you had in mind?

nvaytet (dismissed):

docs/user-guide/common/beam-center-finder.ipynb
SimonHeybrock: #136 will come first. I will update this PR once that is complete.

src/ess/isissans/mantidio.py
```diff
@@ -32,10 +32,10 @@ def __getattr__(self, name: str) -> NoReturn:
     # Needed for type annotations
     MatrixWorkspace = object
 
-Period = NewType('Period', int)
+Period = NewType('Period', int | None)
```
SimonHeybrock, line 35: I was not sure if the `Optional` should be absorbed into the aliases like this. As it is now it is less explicit on the user side, but requires less changes. If we don't like it we can change later on?

docs/user-guide/loki/loki-direct-beam.ipynb
```diff
@@ -558,11 +556,8 @@
     "workflow[DimsToKeep] = []\n",
     "\n",
     "# Set parameter series for file names\n",
```
nvaytet, line 558: Outdated comment?

docs/user-guide/isis/zoom.ipynb
```diff
@@ -140,14 +140,16 @@
...
+    "\n",
     "workflow[DirectBeamFilename] = isis.data.zoom_tutorial_direct_beam()\n",
     "workflow[isis.CalibrationFilename] = isis.data.zoom_tutorial_calibration()\n",
     "workflow[Filename[SampleRun]] = isis.data.zoom_tutorial_sample_run()\n",
     "workflow[Filename[EmptyBeamRun]] = isis.data.zoom_tutorial_empty_beam_run()\n",
     "workflow[isis.SampleOffset] = sc.vector([0.0, 0.0, 0.11], unit='m')\n",
     "workflow[isis.DetectorBankOffset] = sc.vector([0.0, 0.0, 0.5], unit='m')\n",
     "masks = isis.data.zoom_tutorial_mask_filenames()\n",
-    "workflow.set_param_series(PixelMaskFilename, masks)"
+    "workflow = set_pixel_mask_filenames(workflow, masks)"
```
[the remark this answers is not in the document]
nvaytet: I guess when all this will be hidden inside small widgets, it won't be very important anymore...

src/ess/isissans/mantidio.py
nvaytet: I prefer it like this, never really liked the `Optional` thing anyway

src/ess/sans/masking.py
```diff
@@ -4,24 +4,54 @@
...
+def get_detector_ids_from_sample_run(data: TofData[SampleRun]) -> DetectorIDs:
+    """Extract detector IDs from sample run."""
+    return DetectorIDs(
+        data.coords[
+            'detector_number' if 'detector_number' in data.coords else 'detector_id'
+        ]
+    )
+
+
+def to_detector_mask(
```
nvaytet, line 30: This makes more sense :+1:

nvaytet (commented):
DId you verify that the results are identical to what you get from `main`?

docs/user-guide/loki/loki-direct-beam.ipynb
```diff
@@ -308,6 +305,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "workflow = sans.set_banks(workflow, banks=['larmor_detector'])\n",
```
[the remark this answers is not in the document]
SimonHeybrock: No, I think I put it there having the full Loki in mind, to make it easier to modify in the future. Should I remove it?

docs/user-guide/isis/zoom.ipynb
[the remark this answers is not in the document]
SimonHeybrock: Alternatives are:
- Make this arguments of the workflow creation, `LokiWorkflow(sample_runs=..., ...)`.
- Turn `LokiWorkflow` into a builder (using builder pattern):
  ```python
  workflow = LokiWorkflow()
  workflow.set_sample_runs([...])
  workflow.build()  # or build automatically in `workflow.compute()`, by making this a wrapper.
- Turn `LokiWorkflow` into a wrapper, with internal reference to the pipeline. Dedicated methods can set and update the internal workflow.

I think neither of those options is too crazy, but we should probably give it more thought, and gain some more experience using this.

I opened scipp/essreduce#35 for tracking this.

SimonHeybrock:
> DId you verify that the results are identical to what you get from `main`?

Only by eye, looking at the figures in the notebooks.

docs/user-guide/loki/loki-direct-beam.ipynb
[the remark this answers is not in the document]
nvaytet: I would say yes

nvaytet:
> Only by eye, looking at the figures in the notebooks.

I think it would be good to check the values.

SimonHeybrock:
> > Only by eye, looking at the figures in the notebooks.
> 
> I think it would be good to check the values.

I updated the Loki I(Q) test to verify against reference files. I created those by running the same test on `main`. The files have been added to this repo in this branch. I excluded the "upper-bound" mode for now, since we expect changes there soon (masking?), which would break the test. Since this PR is unrelated to the uncertainty broadcast mode I think we can just add those reference files later, if desired.

docs/api-reference/index.md
```diff
@@ -19,6 +19,10 @@
    :recursive:
 
    providers
+   set_background_runs
```
nvaytet, line 22: As was suggested in https://github.com/scipp/essdiffraction/pull/63, should we rename to `with_*`?

Otherwise, everything looks good :+1:

SimonHeybrock: Done

SimonHeybrock:
I updated to the released Sciline. I believe all comments have been addressed so I will merge and release once the builds pass.

PR #136, opened 2024-05-01
UX: Improve readability for ISIS workflows

SimonHeybrock:
Implements changes as suggested for guideline addition in scipp/essreduce#22.

This goes one step further and also improves the currently very confusing way of switching between tutorial data from pooch and "local" data.


docs/user-guide/isis/sans2d.ipynb
```diff
@@ -47,97 +48,129 @@
...
    "metadata": {},
    "outputs": [],
    "source": [
-    "# left-right layout works better for this graph\n",
-    "pipeline.visualize(BackgroundSubtractedIofQ, graph_attr={'rankdir': 'LR'})"
+    "workflow[DataFolder] = isis.data.get_sans2d_tutorial_data_folder()\n",
+    "workflow[DirectBeamFilename] = 'DIRECT_SANS2D_REAR_34327_4m_8mm_16Feb16.dat.h5'\n",
+    "workflow[Filename[SampleRun]] = 'SANS2D00063114.nxs.h5'\n",
+    "workflow[Filename[BackgroundRun]] = 'SANS2D00063159.nxs.h5'\n",
+    "workflow[Filename[EmptyBeamRun]] = 'SANS2D00063091.nxs.h5'"
```
jl-wynen, line 165: I'm not convinced by this approach. It is nice that `get_sans2d_tutorial_data_folder` places all files in the same folder and that we can then simply list the file names. But this pretends that the mechanism is flexible while in reality, the filenames are determined by the `get_*_data_folder` functions. So there is no need to list them separately.

If you want to use this as an example for users, this doesn't quite work either. Users would typically (I think) not provide a folder with all their files and list `Filename`s. Instead, they would, if they use local files directly, list `FilePath`s.

With SciCat, it would also not be possible to list file names after downloading because the names are required for the download itself.

src/ess/isissans/sans2d.py
```diff
@@ -66,3 +72,25 @@ def mask_detectors(
...
+    """
+    Create Sans2d tutorial workflow.
+
+    Equivalent to :func:`Sans2dWorkflow`, but with loaders for tutorial data instead
+    of Mantid-based loaders.
+    """
+    workflow = Sans2dWorkflow()
+    workflow.insert(load_tutorial_run)
+    workflow.insert(load_tutorial_direct_beam)
+    return workflow
```
jl-wynen, line 96: Does it really make sense to hide the insertions in this function? It is convenient for the tutorials but I see 2 issues with it:
- It increases the number of functions that a user needs to know about.
- It is unclear what is different about this tutorial workflow compared to the regular one. Users may assume that there are more fundamental differences than just how files are loaded.

docs/user-guide/isis/sans2d.ipynb
SimonHeybrock: > If you want to use this as an example for users, this doesn't quite work either. Users would typically (I think) not provide a folder with all their files and list Filenames. Instead, they would, if they use local files directly, list FilePaths.

Yes, this was intended as an example for users. But this mechanism is what we have previously implemented, using the `DataFolder` mechanism. They can just set the to `'.'`, no?

src/ess/isissans/sans2d.py
SimonHeybrock: - Why does it increase the number? It decreases it from 2 (`load_*`) to 1 (`Sans2dTutorialWorkflow`).
- Why is it unclear? It is in the docstring as well as mentioned in the notebook.

docs/user-guide/isis/sans2d.ipynb
SimonHeybrock: > With SciCat, it would also not be possible to list file names after downloading because the names are required for the download itself.

Isn't this a different purpose?
- For SciCat download, the user needs to list which files they need.
- For the workflow, they need to specify which file has which meaning (sample, background, ...).

So it is not clear to me if that duplication can be avoided? Even though one could imagine that download all files from an experiment, such that there is no need to list them individually for download?

jl-wynen: > Yes, this was intended as an example for users. But this mechanism is what we have previously implemented, using the DataFolder mechanism. They can just set the to '.', no?

Only if file paths are specified relative to CWD. And then it still goes through the path resolution `Filename` -> `FilePath`. I think it is cleanest to specify a `FilePath` directly to avoid any potential issues with having a bad resolver.


> Isn't this a different purpose?
> 
>     * For SciCat download, the user needs to list which files they need.
> 
>     * For the workflow, they need to specify which file has which meaning (sample, background, ...).
> 
> 
> So it is not clear to me if that duplication can be avoided? Even though one could imagine that download all files from an experiment, such that there is no need to list them individually for download?

The datasets should, at least partially, indicate the meaning of a file. E.g., a sample dataset should link to a background dataset, explicitly stating that this is a background. See https://github.com/scipp/essreduce/blob/5f23eaca66727f2507eb0d89896a947d106029d0/src/ess/reduce/scicat.py#L29 as proposed in https://github.com/scipp/essreduce/pull/18.

And conceptually, we also have a (`Filename`, `DatasetId`) -> `FilePath` resolver. So it is pretty much the same as in the pooch case except that it needs an additional input.

src/ess/isissans/sans2d.py
jl-wynen: > - Why does it increase the number? It decreases it from 2 (load_*) to 1 (Sans2dTutorialWorkflow).

ok

> - Why is it unclear? It is in the docstring as well as mentioned in the notebook (even though that could be extended).

The explanation in the notebook is not completely clear. And many users won't find the docstring. I would like the explanation to be clear about the fact that the workflows are identical except for how they read files and that the files are entirely equivalent and just use a different format.

docs/user-guide/isis/sans2d.ipynb
SimonHeybrock: ### Conclusion from discussion:

We are considering to remove `DataFolder` and `FilePath`. These were previously added to "hide" differences between local files and files obtained via Pooch or Scicat. However, they are confusing to users, and complicate the visualized task graph.
Instead, use something like:

```python
# Note that this assumes the helper functions return sciline.Pipeline, which Pipeline.__setitem__
# can handle after the rewrite. Otherwise use `workflow.insert`.
workflow[Filename[SampleRun]] = download_tutorial_sample_data()
workflow[Filename[SampleRun]] = download_scicat_dataset(12345)
```

There are variations to this where a SciCat helper finds, e.g., background runs related to the specified sample run, such that those can be inserted into the workflow automatically.

There are some related questions around other meta data in the SciCat dataset that may be needed by the workflow. There are several solutions to this, but none of them appears to interfere with the decision to remove `DataFolder` and `FilePath`.

SimonHeybrock: @nvaytet Thoughts?

src/ess/isissans/data.py
```diff
@@ -94,22 +71,54 @@ class IsisRegistry:
...
+def get_sans2d_tutorial_data_folder() -> DataFolder:
+    """Get the path to the folder containing the SANS2D tutorial data."""
+    for filename in (
+        'DIRECT_SANS2D_REAR_34327_4m_8mm_16Feb16.dat.h5',
+        'SANS2D00063091.nxs.h5',
+        'SANS2D00063114.nxs.h5',
+        'SANS2D00063159.nxs.h5',
+    ):
+        path = get_path(filename)
+    return DataFolder(str(Path(path).parent))
```
nvaytet, line 89: Is this just returning the path of the last file? Maybe that was intentional? Should we check the paths are all the same for all files?

docs/user-guide/isis/sans2d.ipynb
```diff
@@ -377,100 +410,6 @@
    "source": [
     "pp.plot(sc.collapse(result.hist(), keep='Q'), norm='log')"
    ]
-  },
-  {
-   "cell_type": "markdown",
-   "id": "0977a3dc-eb5f-4067-80c5-cc19137cc915",
-   "metadata": {},
-   "source": [
-    "## Loading local files\n",
```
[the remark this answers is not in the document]
SimonHeybrock: See above, I think we will remove `DataFolder` and `FilePath`, as it is not needed any more and too confusing.

docs/user-guide/isis/zoom.ipynb
```diff
@@ -43,168 +35,156 @@
...
-    "    'left_beg_18_2.xml',\n",
-    "    'right_beg_18_2.xml',\n",
-    "    'small_bs_232.xml',\n",
-    "    'small_BS_31032023.xml',\n",
-    "    'tube_1120_bottom.xml',\n",
-    "    'tubes_beg_18_2.xml',\n",
-    "]"
+    "workflow = isis.zoom.ZoomTutorialWorkflow()\n",
+    "# For real data use:\n",
+    "# workflow = isis.zoom.ZoomWorkflow()"
```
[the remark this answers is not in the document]
SimonHeybrock: I see nothing wrong with the repetition, maybe we will find it useful in the name. Also, otherwise you have multiple identically named functions or classes in different submodules of the same package.

src/ess/isissans/data.py
SimonHeybrock: If we change as discussed above this would become irrelevant.

docs/user-guide/isis/sans2d.ipynb
nvaytet: I think it's great if we can remove `DataFolder` and `FilePath`.
I was wondering if the above could still be a bit confusing. If you read it twice, and are used to it, then it makes sense that `download_tutorial_sample_data()` would download the file somewhere, and then return the path to that file.
But the first time I read it, I sort of expected `download_tutorial_sample_data()` to return the actual data, not the `Filename`.
Can a slighty different (but not too lengthy) name help?

src/ess/isissans/sans2d.py
SimonHeybrock: Agreed to add an extra sentence in notebook to address this.

SimonHeybrock:
I think I address all comments now. `FilePath` and `DataFolder` are gone. Ready for another review!

docs/user-guide/isis/sans2d.ipynb
```diff
@@ -47,97 +49,130 @@
...
    "cell_type": "code",
    "execution_count": null,
-   "id": "d71780b8-56d5-445f-9d43-635d3d5f406b",
+   "id": "ff1658c5",
    "metadata": {},
    "outputs": [],
    "source": [
-    "# left-right layout works better for this graph\n",
-    "pipeline.visualize(BackgroundSubtractedIofQ, graph_attr={'rankdir': 'LR'})"
+    "workflow[DirectBeamFilename] = isis.data.sans2d_tutorial_direct_beam()\n",
```
nvaytet, line 164: I'm still not entirely sure about this change. I think it may be making the notebook less clear.
Our initial users will be ISs and IDSs, and I think they liked to be able to see the name of the file, because it's usually a file they gave us, and they can then know exactly what is being loaded (even if the file names were not the best).
In addition, it won't necessarily be clear to them what to do if they just want to load a local disk file.

It is also a bit annoying for us that we have to write a function for each file.

Could we make something in-between, like an object or a function that returns file paths? e.g.
```Py
finder = TutorialFileFinder()
workflow[DirectBeamFilename] = finder('DIRECT_SANS2D_REAR_34327_4m_8mm_16Feb16.dat')
workflow[Filename[SampleRun]] = finder('SANS2D00063114.nxs')
...
```
You could then have a `ScicatFileFinder` or something with a better name?

docs/user-guide/isis/zoom.ipynb
```diff
@@ -43,168 +35,150 @@
...
-    "\n",
-    "    for provider in isis.data.providers:\n",
-    "        pipeline.insert(provider)"
+    "workflow[DirectBeamFilename] = isis.data.zoom_tutorial_direct_beam()\n",
+    "workflow[isis.CalibrationFilename] = isis.data.zoom_tutorial_calibration()\n",
+    "workflow[Filename[SampleRun]] = isis.data.zoom_tutorial_sample_run()\n",
+    "workflow[Filename[EmptyBeamRun]] = isis.data.zoom_tutorial_empty_beam_run()\n",
+    "workflow[isis.SampleOffset] = sc.vector([0.0, 0.0, 0.11], unit='m')\n",
+    "workflow[isis.DetectorBankOffset] = sc.vector([0.0, 0.0, 0.5], unit='m')\n",
+    "masks = isis.data.zoom_tutorial_mask_filenames()\n",
```
nvaytet, line 149: For the same reason as above, I'm not sure I like hiding the file names.

docs/user-guide/isis/sans2d.ipynb
SimonHeybrock: > Could we make something in-between, like an object or a function that returns file paths?

That worked out badly so far:
- We had to have special filename mappings (adding suffix, special handling of zip files)
- It is confusing for users, since they may thing that they can use the `finder` for their own files.

> In addition, it won't necessarily be clear to them what to do if they just want to load a local disk file.

I would argue against that. It think it is much clearer now than before, and having `finder` would imho just add more confusion.

> It is also a bit annoying for us that we have to write a function for each file.

I think this is actually a positive change. It avoid hard-coding filenames in notebooks and unit tests.

nvaytet: > since they may thing that they can use the finder for their own files.

True

> It avoid hard-coding filenames in notebooks and unit tests.

I agree that it means if we want to change the file used in tests and the notebooks, we only have one place to change.
However, does this mean we should avoid putting file numbers in the name of the function then? e.g. `loki_tutorial_sample_run_60339` as this would mean if we change that file we have to go and change multiple places in the codebase?
If so, we may struggle to come up with names for all these functions?

SimonHeybrock: > as this would mean if we change that file we have to go and change multiple places in the codebase?

A modern IDE can easily rename such functions (easier than finding all occurrences of a filename)?

I chose to include the run numbers in the Loki case (and only there), since the scientists are actually working with those files, and there are multiple. We can name them `sample_run_1` and `sample_run_2` if you prefer.

src/ess/loki/data.py
```diff
@@ -31,9 +37,48 @@
...
+
+
+def loki_tutorial_sample_transmission_run() -> Filename[TransmissionRun[SampleRun]]:
+    """Sample transmission run (sample + sample holder/can + transmission monitor)."""
+    return Filename[TransmissionRun[SampleRun]](
+        _registry.get_path('60394-2022-02-28_2215.nxs')
+    )
+
+
+def loki_tutorial_run_60392() -> Filename[TransmissionRun[BackgroundRun]]:
```
nvaytet, line 67:
```suggestion
def loki_tutorial_background_transmission_run() -> Filename[TransmissionRun[BackgroundRun]]:
```

docs/user-guide/isis/sans2d.ipynb
nvaytet: I don't really mind either way.
Maybe it's not really that we need to change multiple places, but it would prevent having a function named `loki_tutorial_sample_run_60339()` whose contents get edited and it now returns the file `60777` if one forgets to rename the function.
But once again, I'm ok with either solution.

src/ess/loki/data.py
nvaytet: Maybe we can just also have an `loki_tutorial_empty_beam_run()` function that returns the same file name, which could then be changed to return something else in the future if we need that?

SimonHeybrock: Then the user-visible code is "lying", since it will be invisible that the same run is used. I can change it if you prefer?

docs/user-guide/isis/sans2d.ipynb
SimonHeybrock: I'll leave it then.

src/ess/loki/data.py
nvaytet: It's fine, you can leave it as is.

nvaytet (approved):

PR #143, opened 2024-06-04
Use gravity-corrected two_theta from ScippNeutron

### LINKED ISSUE #90: Move coord transform code from ESSsans to ScippNeutron?
When doing so, make sure to follow the good practices in ScippNeutron regarding preserving precision of inputs. For example, a float32 `tof` will result in float32 `wavelength` or `Q`. I believe the current implementation of conversions with gravity here do not handle this appropriately, i.e., we unintentionally convert to float64 in several places.

jl-wynen:
Fixes #90.

This uses the new functions introduced in https://github.com/scipp/scippneutron/pull/514. It will only work once those are released in ScippNeutron.


SimonHeybrock:
I presume the reference results had to be changed due to the fixed bug?

jl-wynen:
> I presume the reference results had to be changed due to the fixed bug?

Correct

nvaytet (approved):

PR #150, opened 2024-06-26
Do not merge banks

SimonHeybrock:
Based on comment from Andrew Jackson:

> Detector banks at different distances will have to be combined in Q space after averaging to 1D as there will be quite different Q resolutions for the same Q from different distances. Detector panels that are “identical” e.g. left and right on the front array could be merged to a single I(Q) at the event level. Maybe even all 4 front panels in the same bank.

The merge implemented previously is thus not what is required. For now I did not implement other (later in the workflow) merge functionality, I think we should just wait until we get a concrete requirement for this.

Note: PR will fail until we made a bugfix release of Sciline, since `sciline.compute_mapped` raises in the current release.


src/ess/sans/workflow.py
```diff
@@ -52,27 +52,31 @@ def with_pixel_mask_filenames(
...
-def with_banks(workflow: sciline.Pipeline, banks: Iterable[str]) -> sciline.Pipeline:
+def with_banks(
+    workflow: sciline.Pipeline,
+    banks: Iterable[str],
+    index: Iterable[Hashable] | None = None,
+) -> sciline.Pipeline:
     """
     Return modified workflow with bank names set.
 
+    Since banks typically have different Q-resolution the I(Q) of banks are not merged.
```
nvaytet, line 63: So does this mean that when we merge from multiple runs, we need to check that the detector banks were in the same position? The rear panel on Loki can move quite a lot...

SimonHeybrock: ... and we may also need to split runs, if the sample env params are different. There will be more work to do here in the future, I think.

SimonHeybrock:
Updated Sciline, should be ready now.

nvaytet (approved):

PR #153, opened 2024-07-24
Split load of detector/monitor from event-data load

### LINKED ISSUE #39: Should beam center finder construct inner IofQ pipeline based on outer pipeline?
In  [the beam center_finder](https://github.com/scipp/esssans/blob/cc664ecf4cd6073192cb6a38a191d8c19a9a7df5/src/esssans/beam_center_finder.py#L157) we repeatedly compute IofQ in quadrants. This is done by constructing a sciline pipeline and calling `pipeline.compute(IofQ[...])`. This "inner" pipeline is a subgraph of "outer" pipeline.

Right now we construct this subgraph manually inside the beam center finder. I think this is both a bit cumbersome and error-prone. It would be better to take the outer graph, copy it and replace parts of it.
Then we would only need to change `MaskedData[SampleRun]]`, `NormWavelengthTerm[SampleRun]` and `BeamCenter` inside the beam center finder routine, instead of manually copying all required providers to the new pipeline.

Encountered this issue when working on #29 . (See [this](https://github.com/scipp/esssans/pull/29/commits/a88a54ce9073d7deba82019a10d46a614001ed40) for a bad example of how to solve the problem.) In the PR new parameters are added to the IofQ pipeline, for example `DetectorPixelShape[RunType]` per run type. These will also need to be part of the inner pipeline. We can add the new providers as requirements to the beam center finder and add those to the inner pipeline, but it doesn't seem practical if every time we add new components to the outer pipeline we need to remember to also add them in the beam center finder.

I'm not sure what the best solution is here. In principle the inner pipeline should depend on the pipeline it is part of, but maybe that is too complicated. For now it's probably best to just add the new providers as requirements to beam center finder. Maybe this should be converted to an issue in sciline instead.

### LINKED ISSUE #62: Run beam-center finding on current run instead of hard-coding to `SampleRun`
Current implementation:

https://github.com/scipp/esssans/blob/5c239c4ae97efb6a2a2e576eaf9fb8990e05c853/src/esssans/beam_center_finder.py#L275-L285

and

https://github.com/scipp/esssans/blob/5c239c4ae97efb6a2a2e576eaf9fb8990e05c853/src/esssans/beam_center_finder.py#L44-L47

Should we change this to return `BeamCenter[RunType]` and replace `SampleRun` -> `RunType` everywhere?

Points to consider:
- We still want to be able to set the value of the beam center by hand. This would be more complicated if we had to do it for all possible run times. It would be easy to forget some of the run types, leading to inconsistent result.
- We might want to reduce a run that has no sample run. Could be solved by setting beam center by hand.

### LINKED ISSUE #154: Remove on-the-fly beam-center computation?
Currently the SANS workflow can be default compute `BeamCenter` by looking at the center of mass of the sample data. While this is nice, the scientists have previously indicated that this is not how things operate in practice, instead the beam center would be computed once, then reused. The reason I am bringing this up is different though, as explained in the following.

The current approach adds a chain of dependencies in the workflow graph which I fear will cause more trouble in the long run. In simplified terms, the dependency is:
- `BeamCenter->CalibratedData->SolidAngle` &mdash; solid angle computation needs sample positions and detector positions, which we have **after** setting the beam center.
- Because the beam center **can** be computed from detector counts, setting the beam center (producing `CalibratedData`) is done **after** loading event data, **even if the beam center is actually set to a fixed value**.
- An implicit dependency on the loaded masks exists as well.
- This implies that if masks would not be obtained from a file but, e.g., computed based on positions we would have a cyclic dependency (`DetectorMasks->MaskedData->CalibratedData->DetectorMasks`).

The full dependency graph looks as follows (after #153, which made the problem more obvious):

image

An example where this becomes more problematic is chunking processing of files as well as live reduction: The beam center would be set once, but with the current workflow graph we would have to re-compute `SolidAngle` for every chunk. Even when only a single chunk is processed, the graph-level parallelism of the current approach is worse than it has to be.

Another potential example is the TOF computation: It might be that we need the calibrated positions for precise frame-unwrapping, which is currently not possible, since calibration happens after defining TOF.

I therefore propose to re-structure this:

- Pixel calibration should be done *before* loading event data.
- If pixel calibration depends on the event data, I think this would just be run as a separate call to the workflow, i.e., computing the calibration is separate from setting/using it.

This would look as follows:

image

Here`SolidAngle` does not contain pixel masks yet, we would probably add `MaskedSolidAngle` which inserts these. At first I was skeptical about this, since I felt that this would risk "forgetting" to mask the solid angle, which would yield a totally wrong normalization term. However, I now think that it will actually become *more* obvious that we use/need masks in this term by having the explicit `MaskedSolidAngle`. Previously this was kind of implicit and only obvious for an informed reader of the workflow graph (since `SolidAngle` used to depend on `CalibratedMaskedData`, not the "Masked").

Also note #62 which indicates that the current approach may be flawed anyway.

SimonHeybrock:
This addition is related to the new approach discussed in scipp/beamlime#215. ~It depends on a release of scipp/essreduce#53, i.e., CI will fail for now.~

Edit: ~converted to draft, since I found one more dependency on event data that should not be there (for solid angle).~ -> This is fixed by the follow-up #155 (now merged into this branch). -> Fixes #154.
Also closes #39, closes #62.

By loading NXevent_data separately from the containing NXdetector or NXmonitor we increase flexibility, reduce overhead, and continue paving the path towards chunked processing of large files.

Before:

Image

After:

image

Notes:

- By getting the detector IDs from the `RawData[SampleRun]`, which is now the "empty" detector with no events, masks can be loaded and processed without loading potentially huge amounts of events. Even if the events will be loaded anyway this will improve parallelism, and will avoid reprocessing of masks if the event data changes.
- Chunking loading of the event data (using Sciline's map-reduce) with the "after" graph will only load new events, which is also what we need to do for live-reduction. Other detector information as well as masks can be reused. In short, the graph is "wider", which is better.

### DIFF
--- a/docs/user-guide/common/beam-center-finder.ipynb
+++ b/docs/user-guide/common/beam-center-finder.ipynb
@@ -76,7 +76,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "raw = workflow.compute(RawData[SampleRun])['spectrum', :61440]\n",
+    "raw = workflow.compute(RawDetector[SampleRun])['spectrum', :61440]\n",
     "\n",
     "p = isis.plot_flat_detector_xy(raw.hist(), norm='log')\n",
     "p.ax.plot(0, 0, '+', color='k', ms=10)\n",
@@ -128,6 +128,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "workflow[BeamCenter] = sc.vector(value=[0, 0, 0], unit='m')\n",
     "masked = workflow.compute(MaskedData[SampleRun])['spectrum', :61440].copy()\n",
     "masked.masks['low_counts'] = masked.hist().data < sc.scalar(80.0, unit='counts')\n",
     "\n",
@@ -156,9 +157,16 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "# Insert the provider that computes the center-of-mass of the pixels\n",
-    "workflow.insert(sans.beam_center_finder.beam_center_from_center_of_mass)\n",
-    "workflow.visualize(BeamCenter, graph_attr={'rankdir': 'LR'})"
+    "# The center-of-mass approach is based on the MaskedData\n",
+    "workflow.visualize(MaskedData[SampleRun])"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "d7aac9d0",
+   "metadata": {},
+   "source": [
+    "We use the workflow to fill in the required arguments and call the function:"
    ]
   },
   {
@@ -168,7 +176,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "com = workflow.compute(BeamCenter)\n",
+    "com = sans.beam_center_finder.beam_center_from_center_of_mass(workflow)\n",
     "com"
    ]
   },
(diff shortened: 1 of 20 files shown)


SimonHeybrock:
I believe this was reviewed as part of #155.

PR #155, opened 2024-07-25
Workflow dependency restructure

### LINKED ISSUE #154: Remove on-the-fly beam-center computation?

SimonHeybrock:
Fixes #154.

There are some subtleties here, around different way of handling ESS NeXus vs. Mantid files.

The I(Q) beam-center finder was a headache, but I hope by passing the workflow I managed to make it a bit more natural. There is potential room for improvement, by avoiding reloading data even if positions change, but I left it for now until we know if this actually is an issue in practice.

### DIFF
--- a/docs/user-guide/isis/sans2d.ipynb
+++ b/docs/user-guide/isis/sans2d.ipynb
@@ -182,8 +182,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "workflow[BeamCenter] = sc.vector([0, 0, 0], unit='m')\n",
-    "center = workflow.bind_and_call(sans.beam_center_from_center_of_mass)\n",
+    "center = sans.beam_center_from_center_of_mass(workflow)\n",
     "center"
    ]
   },
--- a/docs/user-guide/isis/sans2d.ipynb
+++ b/docs/user-guide/isis/sans2d.ipynb
@@ -72,8 +72,7 @@
    "outputs": [],
    "source": [
     "workflow.insert(isis.data.transmission_from_background_run)\n",
-    "workflow.insert(isis.data.transmission_from_sample_run)\n",
-    "workflow.insert(sans.beam_center_from_center_of_mass)"
+    "workflow.insert(isis.data.transmission_from_sample_run)"
    ]
   },
   {
@@ -165,6 +164,45 @@
     "workflow[Filename[SampleRun]] = isis.data.sans2d_tutorial_sample_run()\n",
     "workflow[Filename[BackgroundRun]] = isis.data.sans2d_tutorial_background_run()\n",
     "workflow[Filename[EmptyBeamRun]] = isis.data.sans2d_tutorial_empty_beam_run()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "981f1f28",
+   "metadata": {},
+   "source": [
+    "The beam center is not set yet.\n",
+    "Unless we have a previously computed value, we can do so now:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "064f87a4",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "workflow[BeamCenter] = sc.vector([0, 0, 0], unit='m')\n",
+    "center = workflow.bind_and_call(sans.beam_center_from_center_of_mass)\n",
+    "center"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "7d7ce61a",
+   "metadata": {},
+   "source": [
+    "Remember to update the workflow with the new center:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "0a4c43f9",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "workflow[BeamCenter] = center"
    ]
   },
   {
@@ -391,7 +429,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.10.12"
+   "version": "3.10.14"
   }
  },
  "nbformat": 4,
--- a/src/ess/isissans/sans2d.py
+++ b/src/ess/isissans/sans2d.py
@@ -5,7 +5,7 @@ from typing import NewType
 import sciline
 import scipp as sc
 from ess.sans import providers as sans_providers
-from ess.sans.types import CalibratedDetector, DetectorMasks, SampleRun
+from ess.sans.types import DetectorMasks, RawDetector, SampleRun
 
 from .data import load_tutorial_direct_beam, load_tutorial_run
 from .general import default_parameters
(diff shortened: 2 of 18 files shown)


src/ess/isissans/sans2d.py
```diff
@@ -22,15 +22,19 @@
 """Sample holder mask"""
 
 
-def detector_edge_mask(sample: TofData[SampleRun]) -> DetectorEdgeMask:
+# It may make more sense to depend on CalibratedDetector here, but the current
```
nvaytet, line 25: Yes, that would make more sense, but we need the masks to be there before we start computing a center of mass, which is annoying.

src/ess/isissans/sans2d.py
```diff
@@ -43,31 +47,31 @@ def sample_holder_mask(
...
-) -> MaskedData[ScatteringRunType]:
-    """Apply pixel-specific masks to raw data.
+def to_detector_masks(
+    edge_mask: DetectorEdgeMask, holder_mask: SampleHolderMask
+) -> DetectorMasks:
+    """Gather detector masks.
+
+    Unlike :py:func:`ess.sans.masking.to_detector_mask`, this function does not rely on
+    mapping using a table of mask filenames. Instead it directly returns a dictionary
+    if multiple masks.
```
nvaytet, line 57: Doesn't it always return a dict of masks, even if there is only one or no mask(s)?

src/ess/sans/beam_center_finder.py
```diff
@@ -171,33 +144,10 @@ def _iofq_in_quadrants(
...
-    params[ElasticCoordTransformGraph] = graph
-    params[BeamCenter] = _offsets_to_vector(data=data, xy=xy, graph=graph)
-    params[DimsToKeep] = ()
-    params[WavelengthMask] = None
-    params[WavelengthBands] = None
-
-    pipeline = sciline.Pipeline(providers, params=params)
-    pipeline[MaskedData[SampleRun]] = data
-    calibrated = pipeline.compute(CalibratedMaskedData[SampleRun])
+    graph = workflow.compute(ElasticCoordTransformGraph)
```
nvaytet, line 147: Nice you could remove all of the above :+1:

docs/user-guide/isis/sans2d.ipynb
```diff
@@ -167,6 +166,45 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "064f87a4",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "workflow[BeamCenter] = sc.vector([0, 0, 0], unit='m')\n",
```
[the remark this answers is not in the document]
SimonHeybrock: Yes. It could in principle also be used to set a default that then gets refined... and I think I should add a test to ensure that we still get the absolute (relative to [0,0,0]) beam center back in a such as case.

src/ess/isissans/sans2d.py
SimonHeybrock:
```suggestion
    of zero or more masks.
```
?

src/ess/isissans/sans2d.py
SimonHeybrock: The masks are anyway applied after setting the beam center (see your other comment https://github.com/scipp/esssans/pull/155#discussion_r1696831864). So this could easily be changed, but maybe that is not how the scientists work? Do you remember where these numbers come from?

SimonHeybrock: Note than in principle we *could* set masks before loading data (when they come from a file).... but maybe that is not desirable if other types of masks come into play?

docs/user-guide/isis/sans2d.ipynb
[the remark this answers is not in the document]
SimonHeybrock: I think this is in fact not working: If we set anything but [0,0,0] the beam-center finder(s) return a beam center relative to the previously set center, which is not what we want, I think?

src/ess/isissans/sans2d.py
nvaytet: Numbers for masks came from trial and error with Wojciech. We just wanted to mask the edges and found values that worked.

src/ess/isissans/sans2d.py
nvaytet: Just "Instead it directly returns a dictionary of masks." ?

docs/user-guide/isis/sans2d.ipynb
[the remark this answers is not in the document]
SimonHeybrock: Please see update, as discussed on the call.

nvaytet (approved):
Looks good to me.
Tests are failing but I assume this is expected?

SimonHeybrock:
> Looks good to me. Tests are failing but I assume this is expected?

Yes, failing due to unreleased and unreviewed changes in ESSreduce.

PR #160, opened 2024-08-20
Use `ess.reduce.nexus` NeXus workflows as basis for all workflows

SimonHeybrock:
I left a couple comments below to explain some of the refactoring.

~CI will fail until ESSreduce is fixed and released.~ (done)


src/ess/isissans/general.py
```diff
@@ -52,45 +74,68 @@ def default_parameters() -> dict:
...
+def to_detector_position_offset(
+    global_offset: DetectorBankOffset, beam_center: BeamCenter
+) -> DetectorPositionOffset[RunType]:
+    return DetectorPositionOffset[RunType](global_offset - beam_center)
+
+
+def to_monitor_position_offset(
+    global_offset: MonitorOffset[MonitorType],
+) -> MonitorPositionOffset[RunType, MonitorType]:
+    return MonitorPositionOffset[RunType, MonitorType](global_offset)
```
SimonHeybrock, line 86: These two helpers replace logic in the removed `components.py`, which was mostly duplicating more generic code.

src/ess/isissans/io.py
```diff
@@ -23,46 +21,6 @@
 CalibrationFilename = NewType('CalibrationFilename', str)
 
 
-def read_xml_detector_masking(filename: PixelMaskFilename) -> MaskedDetectorIDs:
```
SimonHeybrock, line 26: Moved to `ess/sans`: While strictly speaking ISIS-specific, our `ess.loki` unit tests were using this, introducing an annoying dependency of the `ess.loki` unit tests on `ess.isissans`.

src/ess/sans/types.py
```diff
@@ -11,53 +11,36 @@
...
+RunType = reduce_gt.RunType
+SampleRun = reduce_gt.SampleRun
+ScatteringRunType = reduce_gt.ScatteringRunType
+Transmission = reduce_gt.Transmission
+TransmissionRun = reduce_gt.TransmissionRun
+SamplePosition = reduce_gt.SamplePosition
+SourcePosition = reduce_gt.SourcePosition
+
+DetectorBankSizes = reduce_t.DetectorBankSizes
+NeXusDetectorName = reduce_t.NeXusDetectorName
```
SimonHeybrock, line 41: I don't like this, but I think otherwise we would need to rewrite a lot more imports and type hints? This way users can still import from `ess.sans.types` and don't need to know which ones are defined in `ess.reduce.nexus`.... even though that is how the graph viz now displays them :(

SimonHeybrock: I know I am probably alone on this, but see this as another reason for considering #95 again.

nvaytet: Has the situation changed now that we have workflow objects with default parameters set?
I guess the arguments about inspecting intermediate results have not really changed...

src/ess/isissans/general.py
```diff
@@ -52,45 +75,68 @@ def default_parameters() -> dict:
...
-    sample_pos = sample_pos + sample_offset.to(unit=sample_pos.unit, copy=False)
-    pos = data.coords['position']
-    pos = pos + detector_bank_offset.to(unit=pos.unit, copy=False)
-    return RawDetector[RunType](
-        dg['data'].assign_coords(position=pos, sample_position=sample_pos)
-    )
+    # The generic NeXus workflow will try to extract 'data' from this, which is exactly
+    # what we also have in the Mantid data. We use the generic workflow since it also
+    # applies offsets, etc.
+    return NeXusDetector(dg)
```
nvaytet, line 115:
```suggestion
    return NeXusDetector[RunType](dg)
```

src/ess/isissans/general.py
```diff
@@ -52,45 +75,68 @@ def default_parameters() -> dict:
...
 
-def assemble_detector_data(
+def get_monitor_data(
+    dg: LoadedFileContents[RunType], nexus_name: NeXusMonitorName[MonitorType]
+) -> NeXusMonitor[RunType, MonitorType]:
+    # The generic NeXus workflow will try to extract 'data' from this, which is exactly
+    # what we also have in the Mantid data. We use the generic workflow since it also
+    # applies offsets, etc.
+    monitor = dg['monitors'][nexus_name]['data']
+    return NeXusMonitor(sc.DataGroup(data=monitor, position=monitor.coords['position']))
```
nvaytet, line 125:
```suggestion
    return NeXusMonitor[RunType, MonitorType](sc.DataGroup(data=monitor, position=monitor.coords['position']))
```

src/ess/isissans/general.py
```diff
@@ -52,45 +75,68 @@ def default_parameters() -> dict:
...
 ) -> DetectorData[RunType]:
     """Dummy assembly of detector data, detector already contains neutron data."""
     return DetectorData[RunType](detector)
 
 
-def get_monitor_data(
-    dg: LoadedFileContents[RunType], nexus_name: NeXusMonitorName[MonitorType]
-) -> RawMonitor[RunType, MonitorType]:
-    # See https://github.com/scipp/sciline/issues/52 why copy needed
-    mon = dg['monitors'][nexus_name]['data'].copy()
```
nvaytet, line 92: Is the need for this copy now gone? The issue is still open...

src/ess/sans/types.py
SimonHeybrock: You mean the default params we introduced for the widgets? That is not merged and probably still in flux, so I cannot tell yet.

src/ess/loki/general.py
```diff
@@ -6,50 +6,44 @@
...
     TofData,
     TofMonitor,
     TransformationPath,
     Transmission,
     WavelengthBands,
     WavelengthMask,
 )
-from .io import dummy_load_sample
+
+DETECTOR_BANK_RESHAPING = {
```
nvaytet, line 38:
```suggestion
DETECTOR_BANK_SIZES = {
```
(this is the [name used](https://github.com/scipp/essdiffraction/blob/main/src/ess/dream/io/nexus.py#L20) in `essdiffraction`)

src/ess/loki/general.py
```diff
@@ -6,50 +6,44 @@
...
+    'larmor_detector': {'layer': 4, 'tube': 32, 'straw': 7, 'pixel': 512}
+}
 
 
 def default_parameters() -> dict:
     return {
         CorrectForGravity: False,
+        DetectorBankSizes: DETECTOR_BANK_RESHAPING,
```
nvaytet, line 46:
```suggestion
        DetectorBankSizes: DETECTOR_BANK_SIZES,
```

src/ess/loki/general.py
```diff
@@ -205,16 +88,32 @@ def detector_lab_frame_transform(
...
-    get_sample_position,
-    get_source_position,
-    assemble_detector_data,
-    assemble_monitor_data,
     data_to_tof,
     monitor_to_tof,
 )
+
+
+def LokiAtLarmorWorkflow() -> sciline.Pipeline:
```
nvaytet, line 99: Does it make more sense to move this to the `workflow.py` file?
If so, this file would now be pretty empty. Can we remove it altogether? (move stuff to workflow.py or __init__.py?)

src/ess/loki/workflow.py
```diff
@@ -74,30 +67,11 @@ def _build_pipeline(self) -> sciline.Pipeline:
...
-            WavelengthBins: sc.linspace(
-                "wavelength", wavelength_min, wavelength_max, n_wavelength_bins + 1
-            ),
-        }
-        workflow = sciline.Pipeline(providers, params=params)
+        workflow = loki.LokiAtLarmorWorkflow()
+        workflow.insert(_hist_monitor_wavelength)
+        workflow[NeXusMonitorName[Incident]] = "monitor_1"
+        workflow[NeXusMonitorName[Transmission]] = "monitor_2"
+        workflow[WavelengthBins] = sc.linspace("wavelength", 1.0, 13.0, 50 + 1)
```
nvaytet, line 74: :+1:

src/ess/sans/beam_center_finder.py
```diff
@@ -358,29 +361,22 @@ def beam_center_from_iofq(
...
     keys = (
-        RawDetector[SampleRun],
+        NeXusDetector[SampleRun],
         MaskedData[SampleRun],
         NormWavelengthTerm[SampleRun],
         ElasticCoordTransformGraph,
     )
+    workflow = workflow.copy()
+    # Avoid reshape of detector, which would break boolean-indexing by cost function
+    workflow[DetectorBankSizes] = {}
```
nvaytet, line 371: :+1:

tests/loki/iofq_test.py
```diff
@@ -296,8 +277,7 @@ def test_beam_center_from_center_of_mass_is_close_to_verified_result():
 
 
 def test_phi_with_gravity():
-    params = make_params()
-    pipeline = sciline.Pipeline(loki_providers(), params=params)
+    pipeline = make_workflow()
     pipeline[BeamCenter] = _compute_beam_center()
     pipeline[CorrectForGravity] = False
     data_no_grav = pipeline.compute(
```
nvaytet, line 283: I skimmed through the tests, it seemed it was mostly about refactoring by making a fixture that would create the workflows. I don't think there were fundamental changes to any of the tests.
Is that correct?

SimonHeybrock: Correct.

src/ess/isissans/general.py
SimonHeybrock: I think we are not actually accessing a `value` property here?

nvaytet (approved):

PR #161, opened 2024-08-23
deps: add pooch as test dependency in conda build

jokasimr:


YooSunYoung:
https://github.com/scipp/esssans/blob/a246f328eb877de36ec857f9b4fdd303dfda944b/tests/package_test.py#L3-L9

I guess we need to do 3 of them separately...?

jokasimr:
> https://github.com/scipp/esssans/blob/a246f328eb877de36ec857f9b4fdd303dfda944b/tests/package_test.py#L3-L9
> 
> I guess we need to do 3 of them separately...?
> 
> or install `pytest` manually in the action.

Three separately sounds like a good solution

YooSunYoung (approved):

PR #163, opened 2024-08-29
Restructure workflow to facilitate workflow partitioning for efficient streaming execution

SimonHeybrock:
There are two structure changes here:

1. Split the wavelength-reduction (giving the new `ReducedQ`) from the normalization. This is essential for streamed processing since we need to accumulate data (numerator and denominator) before normalization, but as late as possible in the workflow. The wavelength-reduction is relatively expensive since it performs binned data operations,  e.g., to handle wavelength bands. Further, accumulating event data must be avoided since it will quickly run out of memory.
2. Split the detector term from the monitor term in the computation of the denominator. As the detector-term (solid angle and direct beam function) is static, it can be pre-computed. This is important since it involves a relatively costly reduction over many pixels in Q-bins.


src/ess/sans/normalization.py
```diff
@@ -212,22 +213,31 @@ def iofq_norm_wavelength_term(
...
+    direct_beam = direct_beam.transpose(dims)
+    out = solid_angle * broadcast_uncertainties(
+        direct_beam, prototype=solid_angle, mode=uncertainties
+    )
+    # Convert wavelength coordinate to midpoints for future histogramming
+    out.coords['wavelength'] = sc.midpoints(out.coords['wavelength'])
+    return CleanWavelength[ScatteringRunType, Denominator](out)
 
 
 def iofq_denominator(
```
SimonHeybrock, line 239: TODO: This function is unused now, remove it (but keep math in docstring somewhere).

src/ess/sans/conversions.py
```diff
@@ -163,12 +170,44 @@ def detector_to_wavelength(
...
-    return CleanWavelengthMasked[ScatteringRunType, IofQPart](da)
+    return CleanSummedQ[ScatteringRunType, Numerator](da)
+
+
+def mask_wavelength_qxy(
+    da: CleanSummedQxy[ScatteringRunType, Numerator], mask: WavelengthMask
+) -> WavelengthScaledQxy[ScatteringRunType, Numerator]:
+    if mask is not None:
+        da = mask_range(da, mask=mask)
+    return CleanSummedQxy[ScatteringRunType, Numerator](da)
```
nvaytet, line 186:
```suggestion
    return WavelengthScaledQxy[ScatteringRunType, Numerator](da)
```

src/ess/sans/normalization.py
```diff
@@ -168,136 +171,83 @@ def transmission_fraction(
...
+    Compute the monitor-dependent contribution to the denominator term of I(Q).
 
-    Because the multiplication between the ``incident_monitor * transmission_fraction``
-    (pixel-independent) and the direct beam (potentially pixel-dependent) consists of a
-    broadcast operation which would introduce correlations, variances of the direct
-    beam are dropped or replaced by an upper-bound estimation, depending on the
-    configured mode.
+    This is basically ``incident_monitor * transmission_fraction``.
+
+    Keeping the monitor term separate from the detector term allows us to compute the
```
nvaytet, line 183:
```suggestion
    Keeping the monitor term separate from the detector term allows us to compute
```

src/ess/sans/normalization.py
```diff
@@ -168,136 +171,83 @@ def transmission_fraction(
...
+    If the direct beam is not supplied, it is assumed to be 1.
 
-    The denominator is then simply:
-    :math:`M_{\\lambda} T_{\\lambda} D_{\\lambda} \\Omega_{R}`,
-    which is equivalent to ``wavelength_term * solid_angle``.
-    The ``wavelength_term`` includes all but the ``solid_angle`` and is computed by
-    :py:func:`iofq_norm_wavelength_term_sample` or
-    :py:func:`iofq_norm_wavelength_term_background`.
+    Keeping the monitor term separate from the detector term allows us to compute the
+    the latter only once when repeatedly processing chunks of events in streamed data
```
nvaytet, line 217: I guess this is assuming that the pixel positions, and thus the solid angle, remain the same throughout the run.
This is fine I think, as we are no longer computing the beam center as part of the workflow.

Thats said, could the beam center be refined as more and more signal is collected?

src/ess/sans/normalization.py
```diff
@@ -341,13 +291,50 @@ def _normalize(
...
+    where :math:`\\eta` and :math:`\\eta_{M}` are the detector and monitor
+    efficiencies, respectively.
+
+    Hence, in order to normalize the main detector counts :math:`C`, we need compute the
+    transmission fraction :math:`T(\\lambda)`, the direct beam function
+    :math:`D(\\lambda)` and the solid angle :math:`\\Omega(R)`.
+
+    The denominator is then simply:
+    :math:`M_{\\lambda} T_{\\lambda} D_{\\lambda} \\Omega_{R}`,
+    which is equivalent to ``wavelength_term * solid_angle``.
```
nvaytet, line 333: I think `wavelength_term` here was refering to variable names in the code. Should we update or rewrite so that we don't have to change docstring if we change variable names?

tests/loki/iofq_test.py
```diff
@@ -280,13 +280,13 @@ def test_phi_with_gravity():
...
+        to='pixel'
+    )
     graph_no_grav = pipeline.compute(ElasticCoordTransformGraph)
     pipeline[CorrectForGravity] = True
     data_with_grav = (
-        pipeline.compute(CleanWavelengthMasked[SampleRun, Numerator])
+        pipeline.compute(CleanWavelength[SampleRun, Numerator])
         .flatten(to='pixel')
         .hist(wavelength=sc.linspace('wavelength', 1.0, 12.0, 101, unit='angstrom'))
     )
```
nvaytet, line 292: Is it worth adding a test which sort of simulates the streaming? Basically make two chunks of data that are accumulated? I'm not sure what you would test though? Just that it doesn't fail?

Or does this belong more in the data streaming (i.e. in a different repository)?

src/ess/sans/normalization.py
SimonHeybrock: My current idea is to use the stream-processing feature for component position updates, i.e., one lever higher than the Sciline workflow.

src/ess/sans/normalization.py
```diff
@@ -341,13 +291,50 @@ def _normalize(
...
+    The ``wavelength_term`` includes all but the ``solid_angle`` and is computed by
+    :py:func:`iofq_norm_wavelength_term_sample` or
+    :py:func:`iofq_norm_wavelength_term_background`.
```
SimonHeybrock, line 336:
```suggestion
    :math:`M_{\\lambda} T_{\\lambda} D_{\\lambda} \\Omega_{R}`.
```

Maybe removing all references to other code in the docstring is best?

src/ess/sans/normalization.py
SimonHeybrock: How about https://github.com/scipp/esssans/pull/163/files#r1774386408?

tests/loki/iofq_test.py
SimonHeybrock: In any case, it does not belong into this PR. And since even before restructure streaming was possible (it just recomputed more than strictly necessary) it would require mocking some workflow components and adding call counters for the tests. Such tests exist in the `ess.reduce.streaming` module.

nvaytet (approved):

PR #164, opened 2024-08-30
Use new `scipp.bin` and `scipp.hist`

SimonHeybrock:


jl-wynen (approved):
I don't fully understand the old code. But I assume it only did optimisations that are now donw by `bin`/`hist` directly?

SimonHeybrock:
There appears to be an issue where `scipp.hist` drops a coord it should not drop in some case, need to investigate.

PR #167, opened 2024-09-06
Add stream processing examples

SimonHeybrock:
This is not ready for merging yet, but I wanted to share it in case others want to play with the new `ess.reduce.streaming` module. The notebook contains two example, the first simulates event streams, the second uses sliders to select sections from a file.

Needs `essreduce@pre-open-files`


nvaytet:
@SimonHeybrock should we close this? Seems very old and probably no longer relevant?

PR #171, opened 2024-09-27
Workflow widget interface

nvaytet:
Needs https://github.com/scipp/essreduce/pull/109 for builds to pass.


tests/loki/workflow_test.py
```diff
@@ -0,0 +1,102 @@
...
+    # Enable DirectBeam input
+    pbox._input_widgets[DirectBeam].children[0].enabled = True
+    pbox._input_widgets[DirectBeam].children[0].wrapped._option_box.value = None
+    # Run the workflow
+    rbox = wfw.result_box
+    rbox.run_button.click()
+    # Inspect results
+    (da,) = results.values()
+    assert da.dims == ('Q',)
+    assert da.sizes['Q'] == pbox._input_widgets[QBins].children[0].fields['nbins'].value
```
SimonHeybrock, line 102: Won't a test like this break all the time as we refactor and improve the widget code? Can we avoid it (for now at least)?

nvaytet: Sure. Does this mean I should also remove the notebook, as it uses the same code to make the widget come alive?

SimonHeybrock: Let us leave it for now? I hope once we implement a way to save the widget config that would be solved?

PR #172, opened 2024-10-02
Update import locations for new ESSreduce nexus module

SimonHeybrock:
No changes, just updating import location.


src/ess/loki/live.py
jl-wynen, line 1: This file has changed more than just importing things differently. What happened here?

SimonHeybrock: Argh... the follow-up PR was merged into this. Ignore the commits from #173 and have a look again!

jl-wynen (approved):

PR #173, opened 2024-10-02
Update live workflow

SimonHeybrock:
- Use new Protocol defined in beamlime.
- Remove old stateless workflows.
- Assuming renamed entry point (stateless -> workflow_plugin).


YooSunYoung (approved):
I couldn't test it all the way but at least `beamlime` could find the plugin and it also passes the protocol matching test...!

SimonHeybrock:
Ok, let us merge and iterate! No harm can be done as this is not used by anything else.

PR #178, opened 2024-10-08
Update for restructured NeXus base workflow from `ess.reduce.nexus`

### LINKED ISSUE #174: Loki gets lab-frame transform for solid angle from uncalibrated detector position
See https://github.com/scipp/esssans/blob/9434614cf0e2db615779c8d83e375094b059174e/src/ess/loki/general.py#L86-L90
I don't know if this was done correctly in the past and broken in some graph restructure, or if it was always wrong?

As this part of the NeXus workflow is currently being refactored in scipp/essreduce#114 I will fix it as part of those updates.

SimonHeybrock:
Required when scipp/essreduce#114 is released.

Also fixes #174.


nvaytet (approved):
Needs update in dependencies?

PR #181, opened 2024-10-29
fix: don't overwrite input

### LINKED ISSUE #175: Mapping of `event_time_zero` to `tof` modifies input data
Ran into this when working with streaming data. Does not affect "static" workflows.

jokasimr:
Fixes #175 

In the issue it was mentioned that the `isis` module probably had a similar problem. But if that's the case I can't see it.

### DIFF
--- a/src/ess/loki/general.py
+++ b/src/ess/loki/general.py
@@ -61,6 +61,7 @@ def default_parameters() -> dict:
 
 
 def _convert_to_tof(da: sc.DataArray) -> sc.DataArray:
+    da = da.copy(deep=False)
     da.bins.coords['tof'] = da.bins.coords.pop('event_time_offset')
     if 'event_time_zero' in da.dims:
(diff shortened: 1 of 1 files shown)


src/ess/loki/general.py
```diff
@@ -61,6 +61,7 @@ def default_parameters() -> dict:
 
 
 def _convert_to_tof(da: sc.DataArray) -> sc.DataArray:
+    da = da.copy(deep=False)
```
SimonHeybrock, line 64: We do not want to deep-copy since that is costly. We need a shallow copy that also shallow copies the data array representing the bin contents. You can probably use what you implemented in https://github.com/scipp/scipp/pull/3518.

jokasimr: I see what you mean

SimonHeybrock (approved):

PR #182, opened 2024-10-30
Try sharing common parts of workflow compute to make tests run faster

SimonHeybrock:
This is demonstrating an approach that could save some test runtime. In this example a parametrized test is running a nearly identical workflow 8 times, each time loading data from scratch, etc.

The change here moves some of the common parts into a module-scope fixture. On my machine this reduces the test runtime from 50 seconds to 22 seconds.

I do not know if this is a great solution. If we want to go with this, a helper utility could be made: Given a sciline.Pipeline and a set of keys, compute all intermediate results that to not depend on those keys and set the results as "static" data in the workflow. This would avoid error-prone manual authoring of fixtures like the one I added here.

### DIFF
--- a/tests/loki/iofq_test.py
+++ b/tests/loki/iofq_test.py
@@ -13,6 +13,7 @@ import ess.loki.data  # noqa: F401
 from ess import loki, sans
 from ess.sans.conversions import ElasticCoordTransformGraph
 from ess.sans.types import (
+    BackgroundRun,
     BackgroundSubtractedIofQ,
     BackgroundSubtractedIofQxy,
     BeamCenter,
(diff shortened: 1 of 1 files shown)


tests/loki/iofq_test.py
```diff
@@ -98,10 +121,11 @@ def test_pipeline_can_compute_IofQ(uncertainties, qxy: bool):
         BackgroundSubtractedIofQxy,
     ],
 )
-def test_pipeline_can_compute_IofQ_in_event_mode(uncertainties, target):
-    pipeline = make_workflow()
+def test_pipeline_can_compute_IofQ_in_event_mode(
```
nvaytet, line 124: I think the approach is fine. The test looks almost identical to before, it just has a fixture which has some pre-computed stuff in it. It's not making some tests depend on others or anything like that.

How does it change the overall time it takes to run all tests?

SimonHeybrock: > How does it change the overall time it takes to run all tests?

You would have to do something similar for quite a few more tests. But the worse offender is currently the beam-center finder, but I don't feel it is worth playing with that.

SimonHeybrock:
Can we make the files smaller instead? And write more unit tests, instead of testing the full pipeline, where possible?

nvaytet:
> Can we make the files smaller instead? And write more unit tests, instead of testing the full pipeline, where possible?

See an implementation in https://github.com/scipp/essdiffraction/pull/133

nvaytet:
I had made an attempt a while ago to create small files fo tests: https://github.com/scipp/esssans/tree/smaller-test-files
But I never got it working well (although now I can't remember what the issue at the end was...)

PR #184, opened 2024-11-05
Extract common base SansWorkflow and strip NeXus workflow

SimonHeybrock:
- Extract common SansWorkflow
- Use new option of GenericNeXusWorkflow to specify run and monitor types, such that unnecessary nodes can be stripped from the graph.


src/ess/sans/workflow.py
```diff
@@ -131,3 +143,50 @@ def with_background_runs(
 parameter_mappers[NeXusDetectorName] = with_banks
 parameter_mappers[Filename[SampleRun]] = with_sample_runs
 parameter_mappers[Filename[BackgroundRun]] = with_background_runs
+
+
+providers = (
```
SimonHeybrock, line 148: Moved `providers` from `__init__.py` without change.

PR #190, opened 2024-12-16
Limit allowed run and monitor types

jl-wynen:
This reduces the size of workflow graphs in addition to what the arguments to `GenericNeXusWorkflow` accomplish.


SimonHeybrock (approved):
Good catch!

PR #192, opened 2025-01-14
Force alignment on attrs read from file

MridulS:


MridulS:
This should be dealt by https://github.com/scipp/scipp/pull/3632

PR #193, opened 2025-01-17
test: update reference data

### LINKED ISSUE #191: Nightly action failing for Unpinned releases.
Looks like there were updates to scipy optimisers in the [latest 1.15 release](https://docs.scipy.org/doc/scipy-1.15.0/release/1.15.0-notes.html#scipy-optimize-improvements) and this messes up the assertions in one of the tests https://github.com/scipp/esssans/actions/workflows/unpinned.yml.

```
E       AssertionError: 
E       Arrays are not equal
E       when comparing values
E       Mismatched elements: 11 / 100 (11%)
E       Max absolute difference among violations: 4.7683716e-07
E       Max relative difference among violations: 1.1717832e-07
E        ACTUAL: array([4.216311, 4.616472, 4.709669, 4.601788, 4.390278, 4.089487,
E              3.743527, 3.373496, 3.044424, 2.733558, 2.482377, 2.240015,
E              2.070589, 1.893589, 1.749089, 1.633279, 1.532102, 1.421638,...
E        DESIRED: array([4.216311, 4.616472, 4.709668, 4.601788, 4.390278, 4.089487,
E              3.743527, 3.373496, 3.044424, 2.733558, 2.482377, 2.240014,
E              2.070589, 1.893589, 1.749089, 1.633279, 1.532102, 1.421638,...
```

jokasimr:
Fixes #191 

The unpinned tests failed because of changes to `scipy.constants` (specifically the Neutron mass).

I've just replaced the old reference files with what is returned by the workflow.

Do we we want to relax the tolerance on the assertion to avoid getting similar issues in the future?
I imagine a relative tolerance of `1e-5` (currently `1e-7`) might be good enough?


jokasimr:
The test fails because in CI we still have `scipy=1.14.1`

pyproject.toml
```diff
@@ -37,6 +37,7 @@ dependencies = [
   "pandas",
   "plopp",
   "pythreejs",
+  "scipy>=1.15.0",
```
SimonHeybrock, line 40: Hmm, should this pin be only for the tests, maybe? The package works perfectly with older Scipy after all.

jokasimr: That's true :+1:

SimonHeybrock (approved):

jokasimr:
I don't know why this fails. It seems it has something to do with `Optional` / `Switchable` in `essreduce` but I'm not familiar with that. Any ideas?

PR #201, opened 2025-02-28
Test tabular batch widget 'GUI'

olihammond:
This is a very preliminary attempt at having a table-based multiple run reduction using the LoKI workflows and locally stored nxs , mask, and direct beam datafiles. I think there is a _lot_ that can be done to make this 'smarter' and more dynamic by integrating some of the previous workflow widget features. However, this could be seen as a step towards a more representative ess sans "GUI".


olihammond:
> Thanks for bringing a concrete proposal! Is this in any way related to what @jokasimr has put together for reflectometry? Can we look into finding a unified approach?

Excellent question. I'd be very happy to see more unification; certainly the tabular input was strongly inspired by the reflectometry work by @jokasimr. Perhaps there are components both could use from the other, or some smart refactoring and condensation that could be done. I'm not sure how much we (or at least, I and my limited talents) can achieve this, because we run into a lot of technique-specific quirks quite quickly, namely, plotting and the manner of interaction with the underlying Scipp workflows.

PR #206, opened 2025-04-09
Cache paths returned from Pooch to avoid repeated checksum in unit tests

SimonHeybrock:
This seems to save, e.g., 0.3 seconds a whole bunch of times. Not huge but noticeable. Is it a good idea?

### DIFF
--- a/src/ess/sans/data.py
+++ b/src/ess/sans/data.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
+from functools import cache
 
 
 class Registry:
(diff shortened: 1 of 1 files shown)


src/ess/sans/data.py
```diff
@@ -18,6 +19,7 @@ def __init__(self, instrument: str, files: dict[str, str], version: str):
     def __contains__(self, key):
         return key in self._registry.registry
 
+    @cache  # noqa: B019
```
jl-wynen, line 22: Can you put a comment why it is ok to cache `self` in this case?

jl-wynen (commented):
We have this code spread out across most (all?) of our packages. And it is slightly different everywhere. Can we move this to a common location so that we benefit from caching everywhere?

PR #208, opened 2025-04-11
fix: use new Scope

jokasimr:
`Scope` in sciline has recently seen some improvements, and using the old version `ScopeTwoParams` fails the nightly tests on main (because of the deprecation warning.


jl-wynen:
This is a duplicate of #207

PR #212, opened 2025-05-26
Fix run mapping functionality

SimonHeybrock:
Fixes scipp/ess#341.

Note
> But this poses the question why the implementation prior to [#163](https://github.com/scipp/esssans/pull/163) computed the transmission fraction as it does, on a per-run basis. Wouldn't there be less noise (when using transmission from sample run) if we summed each monitor before computing `TransmissionFraction`?

This is not addressed in this PR. Is this something that needs fixing?

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -141,7 +141,7 @@ sciline==25.4.1
     # via
     #   -r base.in
     #   essreduce
-scipp==25.5.0
+scipp==25.5.1
     # via
     #   -r base.in
     #   essreduce


src/ess/sans/workflow.py
```diff
@@ -99,8 +99,8 @@ def _set_runs(
     pipeline = pipeline.copy()
     runs = pd.DataFrame({Filename[key]: runs}).rename_axis(axis_name)
     for part in (Numerator, Denominator):
-        pipeline[CleanSummedQ[key, part]] = (
-            pipeline[CleanSummedQ[key, part]]
+        pipeline[WavelengthScaledQ[key, part]] = (
```
nvaytet, line 102: I guess my only question is: did the results change at all after this modification?

src/ess/isissans/mantidio.py
```diff
@@ -99,12 +100,28 @@ def from_data_workspace(
 
 
 def load_run(filename: Filename[RunType], period: Period) -> DataWorkspace[RunType]:
-    loaded = _mantid_simpleapi.Load(
-        Filename=str(filename), LoadMonitors=True, StoreInADS=False
-    )
+    # Loading many small files with Mantid is, for some reason, very slow when using
+    # the default number of threads in the Dask threaded scheduler (1 thread worked
```
nvaytet, line 104: Can it be because multi-threading is also used internally by Mantid when it is loading the data?

nvaytet (approved):

src/ess/isissans/mantidio.py
SimonHeybrock: I think I checked when I first wrote this (note this code and comment was simply moved in this PR). I don't think so.

PR #214, opened 2025-06-06
Support ISIS histogram files

SimonHeybrock:
What the title says. Needed for processing ZOOM polarized data.

So far I do not have permission to share the files, so there is no added test.

### DIFF
--- a/src/ess/isissans/general.py
+++ b/src/ess/isissans/general.py
@@ -4,7 +4,8 @@
 Providers for the ISIS instruments.
 """
 
-from typing import NewType
+from dataclasses import dataclass
+from typing import Generic, NewType
 
 import sciline
 import scipp as sc
@@ -45,6 +46,17 @@ class MonitorOffset(sciline.Scope[MonitorType, sc.Variable], sc.Variable):
     """
 
 
+@dataclass
+class MonitorSpectrumNumber(Generic[MonitorType]):
+    """
+    Spectrum number for monitor data.
+
+    This is used to identify the monitor in ISIS histogram data files.
+    """
+
+    value: int
+
+
 DetectorBankOffset = NewType('DetectorBankOffset', sc.Variable)
 SampleOffset = NewType('SampleOffset', sc.Variable)
 
@@ -61,6 +73,10 @@ def default_parameters() -> dict:
         SampleOffset: SampleOffset(sc.vector([0, 0, 0], unit='m')),
         NonBackgroundWavelengthRange: None,
         Period: None,
+        # Not used for event files, setting default so the workflow works. If histogram
+        # data is used, the user should set this to the correct value.
+        MonitorSpectrumNumber[Incident]: MonitorSpectrumNumber[Incident](-1),
+        MonitorSpectrumNumber[Transmission]: MonitorSpectrumNumber[Transmission](-1),
     }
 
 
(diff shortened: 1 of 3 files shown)


src/ess/isissans/general.py
```diff
@@ -132,12 +148,43 @@ def get_calibrated_isis_detector(
...
 def get_monitor_data(
-    dg: LoadedFileContents[RunType], nexus_name: NeXusMonitorName[MonitorType]
+    dg: LoadedFileContents[RunType],
+    nexus_name: NeXusMonitorName[MonitorType],
+    spectrum_number: MonitorSpectrumNumber[MonitorType],
 ) -> NeXusComponent[MonitorType, RunType]:
+    """
+    Extract monitor data that was loaded together with detector data.
+
+    If the raw files is histogram data, Mantid stores this as a Workspace2D, where some
```
jl-wynen, line 158:
```suggestion
    If the raw file is histogram data, Mantid stores this as a Workspace2D, where some
```

src/ess/isissans/general.py
```diff
@@ -132,12 +148,43 @@ def get_calibrated_isis_detector(
...
+    or all spectra corresponds to monitors.
```
jl-wynen, line 159:
```suggestion
    or all spectra correspond to monitors.
```

src/ess/isissans/general.py
```diff
@@ -132,12 +148,43 @@ def get_calibrated_isis_detector(
...
+
+    Parameters
+    ----------
+    dg:
+        Data loaded with Mantid and converted to Scipp.
+    nexus_name:
+        Name of the monitor in the NeXus file, e.g. 'incident_monitor' or
+        'transmission_monitor'. USed when raw data is an EventWorkspace, where
```
jl-wynen, line 167:
```suggestion
        'transmission_monitor'. Used when raw data is an EventWorkspace, where
```

src/ess/isissans/general.py
```diff
@@ -132,12 +148,43 @@ def get_calibrated_isis_detector(
...
+        monitors are stored in a group with this name.
+    spectrum_number:
+        Spectrum number of the monitor in the NeXus file, e.g. 0 for incident monitor
+        or 1 for transmission monitor. This is used when the raw data is a
```
jl-wynen, line 171: Are those example numbers correct? The default params are -1. Which also seems odd because both monitors have the same.

src/ess/isissans/mantidio.py
```diff
@@ -99,12 +100,29 @@ def from_data_workspace(
 
 
 def load_run(filename: Filename[RunType], period: Period) -> DataWorkspace[RunType]:
-    loaded = _mantid_simpleapi.Load(
-        Filename=str(filename), LoadMonitors=True, StoreInADS=False
-    )
+    # Loading many small files with Mantid is, for some reason, very slow when using
+    # the default number of threads in the Dask threaded scheduler (1 thread worked
+    # best, 2 is a bit slower but still fast). We can either limit that thread count,
+    # or add a lock here, which is more specific.
```
jl-wynen, line 106: Interesting. This might also help with other loaders. I saw some slowdown in BIFROST when using a threaded scheduler.

src/ess/isissans/mantidio.py
```diff
@@ -99,12 +100,29 @@ def from_data_workspace(
...
+    with load_run.lock:
+        loaded = _mantid_simpleapi.Load(Filename=str(filename), StoreInADS=False)
+    try:
+        loaded = _mantid_simpleapi.Load(
+            Filename=str(filename), LoadMonitors=True, StoreInADS=False
+        )
+    except TypeError:
+        # Not loaded using LoadEventNexus, so LoadMonitor option is not available.
+        loaded = _mantid_simpleapi.Load(Filename=str(filename), StoreInADS=False)
```
jl-wynen, line 115: Why do you call Load again after the lock?

src/ess/isissans/mantidio.py
SimonHeybrock: You think this is due to HDF5?

src/ess/isissans/mantidio.py
SimonHeybrock: Oops, should have been inside the lock. See update,

jl-wynen (approved):

PR #216, opened 2025-07-17
Lower pins

jokasimr:


MridulS:
Updated with dropping python 3.10, copier update and reran tox.

MridulS:
or let it raise the warnings as in https://github.com/scipp/esssans/issues/55

jokasimr:
What remains to be done here?

nvaytet:
> What remains to be done here?

Isn't it just changing the branch protection rules so that the tests it's looking for at the 3.11 ones instead of 3.10?

MridulS:
> I can't see that option in the repo settings, it's probably a permissions issue

Only org owners can do that I believe.

SimonHeybrock:
Done, seems to pass now?

MridulS (approved):

PR #217, opened 2025-07-23
feat: alternative beam center finder

jokasimr:
I was looking at addressing scipp/ess#346 by reducing the runtime of the beam center finder.

As part of that I noticed that the main reason the center of mass beam center finder performs relatively poorly compared to the minimization based approach is the asymmetry of the sample holder mask around the center of mass.

I fixed that by using a symmetrical mask to mask the sample holder.
In the end that method became quite simple, about as fast as the original center of mass approach, and it performs better than the minimization based approach in the sense that it achieves lower cost than the minimization procedure does (surprisingly).

This figure shows what the symmetrical mask looks like:
symmetrical-maks

Note that the center of mass is found iteratively and the mask updates each iteration based on the previous beam center guess.

### Comparison to the minimization based approach

The two figures below show the IoQ curves from the old minimization based approach and the new symmetrical masks + center of mass approach.

old
new

The curve associated with the North East quadrant (NE) is an outlier because it contains the beam holder.
I think that is just because of how this tutorial notebook is written. In a realistic setting the sample holder mask should be taken into account in the normalization factor and then the NE quadrant should not be an outlier any more.


src/ess/sans/beam_center_finder.py
```diff
@@ -37,6 +37,64 @@ def _xy_extrema(pos: sc.Variable) -> sc.Variable:
     return sc.concat([x_min, x_max, y_min, y_max], dim='extremes')
 
 
+def _find_beam_center(
+    data,
+    sample_holder_radius,
+    sample_holder_arm_width,
+):
```
nvaytet, line 44: Could you describe a bit what the algorithm is doing (maybe in a docstring)? It's quite hard to follow, also with the variable names like `c` and `d`.

If I followed a little, it seems this assumes masks will always be wedge-like (because sample holder will be an arm stretching out from one side). I am not sure we can make that assumption in all cases?

src/ess/sans/beam_center_finder.py
```diff
@@ -37,6 +37,64 @@ def _xy_extrema(pos: sc.Variable) -> sc.Variable:
...
+    m = data.copy()
+    m.masks.clear()
+    s = m.bins.sum()
+
+    for i in range(20):
```
nvaytet, line 49: Can you explain where the number 20 (and 10 below) comes from?

jokasimr: Just iteration counts that I made up. You can probably reduce them by a lot but the iterations are quick anyway so it doesn't matter.

src/ess/sans/beam_center_finder.py
jokasimr: I'll update with a description.

`c` is the current beam center guess

`d` is the distance from the current beam center guess to each pixel

jokasimr:
For reference, this is pseudocode for how I imagine we could symmetrize an arbitrary user defined sample holder mask:

```python
def symmetric_mask(center, mask):
   '''Assumes we can compute the center point of any pixel
     and that we can compute the index of the pixel that contains a point.
   '''
    m = mask.copy()
    
    for i in range(mask.size):

        # Pixel position mirrored in center
        im = round(2 * center - position(mask, i))

        # Position to pixel index
        j = position_to_index(mask, im)

        if 0 <= j < m.size and mask[i]:
            m[j] = True
            
    return m
```

nvaytet:
@jokasimr not sure if you saw the review comments above or not?

jokasimr:
As a sanity check I used the alternative center of mass beam finder in the `user-guide/loki/loki-iofq.ipynb` tutorial notebook.
The results are very similar, which is expected since there's no beam stop shadow in the detector images.

deut-SDS
Porous_silica
AgBeh

PR #218, opened 2025-08-01
feat: pooch retry

jokasimr:


src/ess/sans/data.py
```diff
@@ -3,7 +3,9 @@
 
 
 class Registry:
-    def __init__(self, instrument: str, files: dict[str, str], version: str):
+    def __init__(
+        self, instrument: str, files: dict[str, str], version: str, retry_if_failed=3
```
YooSunYoung, line 7:
```suggestion
        self, instrument: str, files: dict[str, str], version: str, retry_if_failed: int=3
```

Just to be consistent with other type hints... : D...

YooSunYoung (approved):

PR #219, opened 2025-08-07
Drop support for Python 3.10

jl-wynen:
And copier update.


jl-wynen:
Never mind, this is now part of #216

PR #246, opened 2026-03-09
Lookup relative error threshold

### LINKED ISSUE #338: Position coord on CorrectedMonitor causes transmission_fraction to fail
## Summary

`transmission_fraction` in `ess.sans.normalization` fails with a `DatasetError` (coordinate mismatch on `position`) when computing the transmission fraction using monitors from the `GenericTofWorkflow` path (i.e., via `LokiWorkflow`).

The root cause: `monitor_to_wavelength` calls `transform_coords('wavelength', graph=graph, keep_inputs=False)`, but `position` is **not an input** to the `MonitorCoordTransformGraph`. The graph starts from pre-computed `Ltotal`, so `position` is a bystander coordinate that survives the conversion. The `transmission_fraction` provider then multiplies ratios from incident and transmission monitors, which sit at different physical positions, causing scipp to reject the operation.

## Reproduction

The error occurs when:
- Monitor data is loaded from a NeXus geometry file (via `assemble_monitor_data`), which attaches `position` from `EmptyMonitor`
- The data flows through `monitor_to_wavelength` → `preprocess_monitor_data` → `CorrectedMonitor`
- `transmission_fraction` computes `(sample_transmission / direct_transmission) * (direct_incident / sample_incident)` — each division preserves the respective monitor's `position`, and the final multiply fails

```
scipp._scipp.core.DatasetError: Mismatch in coordinate 'position' in operation 'multiply':
()    vector3              [m]  (0, 0, 0.2)
vs
()    vector3              [m]  (0, 0, -8.4)
```

## Verification

```python
import scipp as sc

# Simulates what monitor_to_wavelength does
da = sc.DataArray(
    sc.ones(dims=['tof'], shape=[10]),
    coords={
        'tof': sc.linspace('tof', 1000, 50000, 10, unit='us'),
        'position': sc.vector([0, 0, -8.4], unit='m'),
        'Ltotal': sc.scalar(25.0, unit='m'),
    }
)

from scippneutron.conversion import graph as conv_graph
graph = {
    **conv_graph.beamline.beamline(scatter=False),
    **conv_graph.tof.elastic_wavelength('tof'),
    'source_position': lambda: sc.vector([0, 0, -60.0], unit='m'),
}

result = da.transform_coords('wavelength', graph=graph, keep_inputs=False)
print(list(result.coords))  # ['position', 'wavelength'] — position survives!
```

## Impact

This affects any workflow using `GenericTofWorkflow` (e.g., `LokiWorkflow`) that computes `TransmissionFraction` with monitors at different physical positions. The `LokiAtLarmorWorkflow` is unaffected because its custom `larmor_monitor_to_tof` conversion path doesn't attach `position` the same way.

## Possible fixes

- Drop `position` explicitly in `monitor_to_wavelength` after the `transform_coords` call.
- Or include `position` in the `MonitorCoordTransformGraph` so it gets consumed (and thus dropped by `keep_inputs=False`).
- Or drop non-dimension coords in `preprocess_monitor_data` / `transmission_fraction`.

## Workaround

We currently work around this in ESSlivedata by stripping non-wavelength coords when mapping `CorrectedMonitor` values between run types.

jl-wynen:
Adapt the `LookupTableRelativeErrorThreshold` to be defined per component, as per https://github.com/scipp/essreduce/pull/319

Also:
Fixes scipp/ess#338


requirements/base.txt
```diff
@@ -21,25 +21,25 @@ cyclebane==24.10.0
...
+dask==2026.1.2
     # via -r base.in
 decorator==5.2.1
     # via ipython
 dnspython==2.8.0
     # via email-validator
 email-validator==2.3.0
     # via scippneutron
-essreduce==26.1.1
+essreduce==26.3.1
```
nvaytet, line 32: Should this be set as a lower bound in the pyproject.toml?

nvaytet (commented):
I don't understand why CI is failing... a vector coordinate mismatch does not seem to originate from these changes...

requirements/base.txt
jl-wynen: Done

jl-wynen:
> I don't understand why CI is failing... a vector coordinate mismatch does not seem to originate from these changes...

This is related to a change in ESSreduce. Not sure yet which, but the coord transform from time to wavelength no longer removes the position. Do you know why?

nvaytet:
I now tracked it down to https://github.com/scipp/essreduce/pull/302

nvaytet:
My guess is that we now assign `Ltotal` [here](https://github.com/scipp/essreduce/pull/302/changes#diff-b260f1537704bbfa6c2ae71fef27c3d07833609cda38c57dee14337aa879def1R415) and after that, we do 
```Py
da.transform_coords(
        'wavelength', graph={"wavelength": wavelength_from_tof}, keep_intermediate=False
    )
```
[here](https://github.com/scipp/essreduce/pull/302/changes#diff-b260f1537704bbfa6c2ae71fef27c3d07833609cda38c57dee14337aa879def1R543).

Because the Ltotal is already present, the `position` no longer gets consumed (even though we have `keep_intermediate=False`) and so it is still there on the `WavelengthMonitor`?

nvaytet:
Suggestion for a fix: we manually remove the `position` coords here when we compute the transmission fraction.
We can then remove that once https://github.com/scipp/ess/issues/247 is implemented.

nvaytet:
See https://github.com/scipp/ess/issues/338

src/ess/sans/types.py
```diff
@@ -242,12 +244,6 @@ class IntensityQxQy(sciline.Scope[RunType, sc.DataArray], sc.DataArray):
 """I(Qx, Qy) with background (given by I(Qx, Qy) of the background run) subtracted"""
 
 
-class WavelengthMonitor(
```
nvaytet, line 245: Instead, we can just use the `WavelengthMonitor` from `essreduce`.

src/ess/sans/conversions.py
nvaytet, line 182: This provider was not really needed in the first place. We could have just used the one from `essreduce` that provides `WavelengthMonitor`. We now use it to 'fix' the monitors (by stripping the position coord).
The entire provider can be removed once we fix this in essreduce.

src/ess/sans/conversions.py
```diff
@@ -183,9 +183,17 @@ def monitor_to_wavelength(
...
-        monitor.transform_coords('wavelength', graph=graph, keep_inputs=False)
-    )
+    """
+    This is a temporary fix: we need to remove the 'position' coordinate if present.
+    This should be done in essreduce; see https://github.com/scipp/ess/issues/247.
+    Once that is implemented, this provider should be removed and we should instead
+    use the ``WavelengthMonitor`` provided by the generic essreduce workflow.
+    """
+
+    out = monitor.transform_coords('wavelength', graph=graph, keep_intermediate=False)
```
nvaytet, line 193: Note here that we went from `keep_inputs=False` (before) to `keep_intermediate=False` (now).
The input now has both `position` and `Ltotal` coords. With `keep_inputs=False`, the Ltotal would disappear and the position would remain.
Now, `keep_intermediate=False` removes nothing (maybe we don't need the kwarg at all?), and we manually remove `position` to only leave Ltotal at the end.

nvaytet:
@jl-wynen can you take another look to check I didn't mess anything up?

jl-wynen (commented):
Looks good. But I can't approve myself because I opened this PR,

nvaytet (approved):

PR #250, opened 2026-03-25
Tofectomy for esssans

nvaytet:
Needed after https://github.com/scipp/ess/pull/65

See some comments below that walk a little through the more confusing changes.


src/ess/loki/larmor_workflow.py
```diff
@@ -0,0 +1,187 @@
+# SPDX-License-Identifier: BSD-3-Clause
```
nvaytet, line 1: I moved the stuff about larmor into a separate file, to leave the loki workflow cleaner.

src/ess/loki/larmor_workflow.py
```diff
@@ -0,0 +1,187 @@
...
+    source_position: Position[snx.NXsource, RunType],
+    gravity: GravityVector,
+) -> ElasticCoordTransformGraph[RunType]:
+    graph = sans_elastic(
+        correct_for_gravity=correct_for_gravity,
+        sample_position=sample_position,
+        source_position=source_position,
+        gravity=gravity,
+    )
+    return ElasticCoordTransformGraph[RunType]({**graph, **tof.elastic_Q('tof')})
```
nvaytet, line 119: Note that this is new in the workflow submodule (was not in the old `workflow.py` file. We put the functions from the `tof` graph back in, as a tof was recorded in the files at Larmor (short pulse source).

tests/loki/reference_IofQ_UncertaintyBroadcastMode.drop.hdf5
nvaytet, line 1: Updated the reference solutions: the data values and variances have remained identical. However, some coordinates pertaining to positions (L1, source_position, sample_position...) are no longer present in the final result. We have Q and wavelength, and this is probably the way it should be.

If we want to add contributions from different measurements, we should no longer care about these positions once we are in Q-space.

src/ess/sans/normalization.py
```diff
@@ -206,11 +206,11 @@ def norm_monitor_term(
     return MonitorTerm[RunType](out)
 
 
-def norm_detector_term(
+def norm_detector_term_denominator(
     solid_angle: CorrectedDetector[RunType, Denominator],
     direct_beam: CleanDirectBeam,
     uncertainties: UncertaintyBroadcastMode,
-) -> WavelengthDetector[RunType, Denominator]:
+) -> DetectorTerm[RunType, Denominator]:
```
SimonHeybrock, line 213: So in summary we have:
- Old: `RawDetector->TofDetector->CorrectedDetector->WavelengthDetector`
- New: `RawDetector->WavelengthDetector->CorrectedDetector->DetectorTerm`

That is:
- Corrected and Wavelength swapped... which brings it in line with https://scipp.github.io/essreduce/user-guide/reduction-workflow-guidelines.html#c-6-domain-types-for-data-products
- New intermediate `DetectorTerm`

src/ess/sans/workflow.py
```diff
@@ -176,7 +176,7 @@ def SansWorkflow() -> sciline.Pipeline:
     :
         SANS workflow as a sciline.Pipeline
     """
-    workflow = GenericTofWorkflow(
+    workflow = GenericUnwrapWorkflow(
```
SimonHeybrock, line 179: To be honest, I would have just kept the `GenericTofWorkflow` - we are still at a TOF neutron source after all.

src/ess/sans/types.py
```diff
@@ -35,12 +35,11 @@
...
 
-LookupTableRelativeErrorThreshold = tof_t.LookupTableRelativeErrorThreshold
-TofDetector = tof_t.TofDetector
-TofMonitor = tof_t.TofMonitor
-TofLookupTableFilename = tof_t.TofLookupTableFilename
-TofLookupTable = tof_t.TofLookupTable
-WavelengthMonitor = tof_t.WavelengthMonitor
+LookupTableRelativeErrorThreshold = unwrap_t.LookupTableRelativeErrorThreshold
+LookupTableFilename = unwrap_t.LookupTableFilename
+LookupTable = unwrap_t.LookupTable
```
SimonHeybrock, line 40: Seems a bit odd to import these unqualified "LookupTable" types directly into the top-level (types module).

src/ess/sans/workflow.py
nvaytet: I don't really mind, others also [suggested other names](https://github.com/scipp/ess/pull/65#pullrequestreview-3927517958).
But this is quite late at this point. Would have been easier to pick a different name earlier.

I'm thinking maybe we don't need the "Generic" in there if we don't want the name to become too long.
So it could become something like `ComputeWavelengthWorkflow`? Or Just `WavelengthWorkflow`.

I could change the name, and then keep `GenericUnwrapWorkflow` for a while as an alias. And phase it out later once we've updated the other packages, so that we don't break everything again.

src/ess/sans/types.py
nvaytet: So you would just import `unwrap` and then have
```Py
from ess.sans.types import unwrap

unwrap.LookupTable
```
?
I can do that.

SimonHeybrock: Yes.

SimonHeybrock: But maybe that is confusing because our `visualize()` currently strips module names from type names.

src/ess/sans/normalization.py
nvaytet: Yes, that is correct.
Screenshot_20260327_091452

SimonHeybrock (approved):

src/ess/sans/types.py
nvaytet: @jokasimr [suggested](https://github.com/scipp/ess/pull/65#discussion_r2916481807) it should be called `WavelengthLookupTable`.
I argued that `unwrap.LookupTable` was descriptive enough.
But yes, it does just show up as `LookupTable` in the visualization.

I don't mind revisiting the naming.

src/ess/sans/workflow.py
jl-wynen: I don't really care.

> we are still at a TOF neutron source after all.

Does this mean we also need `LokiTofWorkflow` and `OdinTofWorkflow`? I think it's enough that the class is defined in an 'ess' package.

nvaytet: I would keep `LokiWorkflow` and `OdinWorkflow`, as they will always want to compute wavelength.

nvaytet: In any case, the naming cannot be changed in this PR, it needs to be done first in `essreduce`.

